PR #222, opened 2019-07-22
IBP-2792 IBP-2958 IBP-1193 roles permissions

clarysabel:
https://github.com/IntegratedBreedingPlatform/BreedingManager/pull/313
https://github.com/IntegratedBreedingPlatform/Fieldbook/pull/808
https://github.com/IntegratedBreedingPlatform/Commons/pull/227
https://github.com/IntegratedBreedingPlatform/Workbench/pull/280
https://github.com/IntegratedBreedingPlatform/BMSAPI/pull/222
https://github.com/IntegratedBreedingPlatform/Middleware/pull/652

PR #223, opened 2019-07-30
IBP-2784 Deprecate crop users and persons tables

abatac:


src/main/java/org/ibp/api/java/impl/middleware/common/MiddlewareFactory.java
```diff
@@ -318,6 +316,12 @@ public WorkbenchDataManager getWorkbenchDataManager() {
 		return new WorkbenchDataManagerImpl(this.getWorkbenchSessionProvider());
 	}
 
+	@Bean
+	@DependsOn("WORKBENCH_SessionFactory")
+	public UserService getUserService() {
```
nahuel-soldevilla: I'm taking note of some things to review in the roles branch #222

src/main/java/org/ibp/api/java/impl/middleware/dataset/validator/StudyValidator.java
```diff
@@ -55,7 +55,7 @@ public void validate(final Integer studyId, final Boolean shouldBeUnlocked) {
 
 		if (shouldBeUnlocked
 			&& study.isLocked()
-			&& !ObjectUtils.equals(study.getCreatedBy(), String.valueOf(contextUtil.getIbdbUserId(loggedInUser.getUserid())))
+			&& !study.getCreatedBy().equals(loggedInUser.getUserid().toString())
```
nahuel-soldevilla: Interesting, I think I've used `ObjectUtils.equals` here in case `createdBy` was null, but it's deprecated now in favor of `java.util.Objects.equals`, perhaps we should use that instead if the reference can be null, though it's not likely here

PR #228, opened 2019-08-07
IBP-2907 Fix Error in CSV Export

abatac:


src/main/java/org/ibp/api/java/impl/middleware/dataset/DatasetCSVGenerator.java
```diff
@@ -60,7 +63,14 @@ public File generateMultiInstanceFile(final Map<Integer, List<ObservationUnitRow
 	String[] getColumnValues(final ObservationUnitRow row, final List<MeasurementVariable> subObservationSetColumns) {
 		final List<String> values = new LinkedList<>();
 		for (final MeasurementVariable column : subObservationSetColumns) {
-			values.add(row.getVariables().get(column.getName()).getValue());
+			if (row.getEnvironmentVariables().containsKey(column.getName()) && (
+				column.getVariableType().getId() == VariableType.ENVIRONMENT_DETAIL.getId()
+					|| column.getVariableType().getId() == VariableType.STUDY_CONDITION.getId())) {
```
darla-leafnode: Create static list of variable types here for readability

src/main/java/org/ibp/api/rest/labelprinting/SubObservationDatasetLabelPrinting.java
```diff
@@ -337,7 +338,12 @@ public LabelsData getLabelsData(final LabelsGeneratorInput labelsGeneratorInput)
 						row.put(requiredField, this.getSeason(observationUnitRow.getVariables().get(field.getName()).getValue()));
 						continue;
 					}
-					row.put(requiredField, observationUnitRow.getVariables().get(field.getName()).getValue());
+					if (observationUnitRow.getVariables().containsKey(field.getName())) {
```
darla-leafnode: Can't we abstract where the value comes from By providing a method in `ObservationUnitRow `class that gets the value where it finds the column name (either from `variables `or `environmentVariables`)?

src/main/java/org/ibp/api/java/impl/middleware/dataset/DatasetCSVGenerator.java
darla-leafnode: Also here, related to comment in abstracting in `ObservationUnitRow `class where value comes from

src/main/java/org/ibp/api/rest/labelprinting/SubObservationDatasetLabelPrinting.java
abatac: Hi @darla-leafnode, I think that'll work if a variable is expected to be present only in either variables or environmentVariables. But what if they are both present in variables or environmentVariables? Then the code needs to decide where to get the value of based on where a variable is added (VariableType). Just like how it is done here: https://github.com/IntegratedBreedingPlatform/BMSAPI/pull/228#discussion_r313206753

Now for this particular code, there's actually a bug in Create Labels, in which, if a variable is present in both environment and observation level, the system will only display that variable under "Environment" section. As a consequence the variable will only be present in environmentVariables. So right now for this code, we just need to get the variable value either from variables or environmentVariables. This works for now, but this should be fixed soon.

src/main/java/org/ibp/api/java/impl/middleware/dataset/DatasetCSVGenerator.java
abatac: > Create static list of variable types here for readability

This is done.

PR #234, opened 2019-09-02
IBP-3017 Adding validation to block users edition when many users share same person id

clarysabel:
Includes
https://github.com/IntegratedBreedingPlatform/Middleware/pull/680


nahuel-soldevilla:
:+1:

PR #237, opened 2019-09-06
IBP-2946 Fix Import Dataset

abatac:


src/main/java/org/ibp/api/java/impl/middleware/dataset/DatasetExcelGenerator.java
```diff
@@ -84,6 +89,15 @@
 	@Resource
 	private DatasetTypeService datasetTypeService;
 
+	@Resource
+	private LocationDataManager locationDataManager;
```
darla-leafnode: locationDataManager is unused

src/main/java/org/ibp/api/java/impl/middleware/dataset/DatasetExcelGenerator.java
```diff
@@ -84,6 +89,15 @@
...
+
+	@Resource
+	private FieldbookService fieldbookService;
```
darla-leafnode: We shouldn't be using `FieldbookService` in BMSAPI, as much as possible

PR #239, opened 2019-09-23
Add service get all variables for dataset

vmaletta:
Please merge. Related to https://github.com/IntegratedBreedingPlatform/Fieldbook/pull/816


src/main/java/org/ibp/api/java/impl/middleware/dataset/DatasetServiceImpl.java
```diff
@@ -583,4 +588,82 @@ private void mapObservationUnitRows(
 			list.add(observationUnitRow);
 		}
 	}
+
+	@Override
+	public List<MeasurementVariable> getColumns(final int studyId, final int datasetId) {
```
nahuel-soldevilla: It seems this functionality is exactly the same as `org.ibp.api.java.impl.middleware.dataset.DatasetCSVExportServiceImpl#getColumns`
Perhaps we can extract it to a `datasetService.getAllVariables`
We should take any opportunity to refactor dataset services code as it's getting too complex, or we may end up with the same problem as in the old workbook services.

vmaletta: It's the same. I will try to reuse it

PR #242, opened 2019-09-30
Change if trndate from int to Date

vmaletta:
Please merge

### DIFF
--- a/src/main/java/org/ibp/api/java/impl/middleware/inventory/InventoryServiceImpl.java
+++ b/src/main/java/org/ibp/api/java/impl/middleware/inventory/InventoryServiceImpl.java
@@ -1,11 +1,6 @@
 
 package org.ibp.api.java.impl.middleware.inventory;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/ibp/api/java/impl/middleware/inventory/InventoryServiceImpl.java
```diff
@@ -125,15 +126,19 @@ private static Integer getCurrentDateInt() {
 		return Integer.valueOf(new SimpleDateFormat("yyyyMMdd").format(Calendar.getInstance().getTime()));
 	}
 
+	private static Date getCurrentDate() {
```
clarysabel: @vmaletta  in my opinion, there is no need to have a function for getting current date

vmaletta: Ok

PR #250, opened 2019-10-31
IBP-3037 IBP-3135 Upgrade swagger

darla-leafnode:


clarysabel:
Hi @darla-leafnode everything seems ok. 
Just a question, why do we need org.codehaus.woodstox?

grg021:
Hi @clarysabel thanks for asking. While I was doing some trial-and-error to see what version of swagger would work for the current codebase I had to add this to make it run properly. Upon testing though for this version of swagger I think we can remove this already. I'll remove it and have it retested.

PR #251, opened 2019-11-11
IBP-3094 IBP-3095 Add Environment and Iterate Design

abatac:


src/main/java/org/ibp/api/rest/study/StudyInstanceResource.java
```diff
@@ -0,0 +1,52 @@
...
+@Controller
+@RequestMapping("/crops")
+public class StudyInstanceResource {
+
+	@Resource
+	private StudyInstanceService studyInstanceService;
+
+	@ApiOperation(value = "Create new study instance",
+		notes = "Create new study instance")
+	@RequestMapping(value = "/{cropname}/studies/{studyId}/instances/{instanceNumber}", method = RequestMethod.PUT)
```
darla-leafnode: We don't need to expose instanceNumber as parameter since it should always just append to the last existing instance
We can make this as POST request and the URL like /{cropname}/studies/{studyId}/instances/generation

src/main/resources/messages_en.properties
```diff
@@ -122,6 +122,7 @@ invalid.observation.id=Invalid observation id.
 invalid.observation.value=Invalid value.
 dataset.name.empty.name=Dataset name should not be empty.
 instance.does.not.exist=Instance does not exist.
+instance.already.exists=Instance number {0} already exists.
```
darla-leafnode: We won't need this validation when we don't expose the instance number as parameter

src/main/java/org/ibp/api/java/impl/middleware/dataset/validator/InstanceValidator.java
```diff
@@ -1,36 +1,104 @@
...
+		if (!instanceNumbers.stream()
+			.distinct()
+			.filter(instancesWithDesign::contains)
+			.collect(Collectors.toSet()).isEmpty() && this.middlewareStudyService.hasAdvancedOrCrossesList(studyId)) {
+			this.errors.reject("study.has.advance.or.cross.list");
+			throw new ApiRequestValidationException(this.errors.getAllErrors());
+		}
+	}
+
+	public void checkStudyInstanceAlreadyExists(final Integer studyId, final int instanceNumber) {
```
darla-leafnode: Same, we don't need this validation if we don't expose the instance number as parameter

src/main/java/org/ibp/api/java/impl/middleware/study/StudyInstanceServiceImpl.java
```diff
@@ -0,0 +1,78 @@
...
+	private DatasetValidator datasetValidator;
+
+	@Resource
+	private InstanceValidator instanceValidator;
+
+	@Override
+	public StudyInstance createStudyInstance(final String cropName, final int studyId, final int instanceNumber) {
+
+		this.studyValidator.validate(studyId, true);
+		this.instanceValidator.checkStudyInstanceAlreadyExists(studyId, instanceNumber);
```
darla-leafnode: Here as well, remove instanceNumber

src/main/java/org/ibp/api/java/impl/middleware/study/StudyInstanceServiceImpl.java
```diff
@@ -0,0 +1,78 @@
...
+	private WorkbenchDataManager workbenchDataManager;
+
+	@Resource
+	private DatasetService datasetService;
+
+	@Resource
+	private StudyValidator studyValidator;
+
+	@Resource
+	private DatasetValidator datasetValidator;
```
darla-leafnode: Seems `datasetValidator` is not needed

src/main/java/org/ibp/api/rest/study/StudyInstanceResource.java
abatac: Done

src/main/resources/messages_en.properties
abatac: Done

src/main/java/org/ibp/api/java/impl/middleware/dataset/validator/InstanceValidator.java
abatac: Done

src/main/java/org/ibp/api/java/impl/middleware/study/StudyInstanceServiceImpl.java
abatac: Done

src/main/java/org/ibp/api/java/impl/middleware/study/StudyInstanceServiceImpl.java
abatac: Done

abatac:
> I think it would be better if we leave the resolving of the next instance number to Middleware. That is, we don't expose it as a parameter in the Middleware service as well.

This is done.

PR #252, opened 2019-11-11
Ibp 3146 lot UUID

vmaletta:
Please, merge.


src/main/java/org/ibp/api/java/impl/middleware/inventory/InventoryServiceImpl.java
```diff
@@ -38,6 +43,11 @@
 	@Autowired
 	private LocationDataManager locationDataManager;
 
+	private LotIDGenerator lotIDGenerator;
```
clarysabel: @vmaletta must be autowired

clarysabel: @vmaletta any advantage of having a generator class instead of handling it in the LotService or DataManager?

vmaletta: Autowired didn't work. 
Based on ObservationUnitIdGenerator, I've created separate class

PR #270, opened 2019-12-16
Add validation for observation dataset

vmaletta:
Please merge


src/main/java/org/ibp/api/java/impl/middleware/dataset/validator/DatasetValidator.java
```diff
@@ -49,17 +49,18 @@ public DatasetValidator() {
 		// Empty constructor expected by MockMvcRequestBuilder
 	}
 
-	public void validateDataset(final Integer studyId, final Integer datasetId, final Boolean shouldBeSubobservationDataset) {
+	public void validateDataset(final Integer studyId, final Integer datasetId, final Boolean shouldBeObservationOrSubobservationDataset) {
```
clarysabel: @vmaletta the variable name is very long. Was trying to find a generic name for both and shouldBeObservationDataset should work

src/main/java/org/ibp/api/java/impl/middleware/dataset/validator/DatasetValidator.java
```diff
@@ -49,17 +49,18 @@ public DatasetValidator() {
...
 			this.errors.reject("dataset.required", "");
 			throw new ApiRequestValidationException(this.errors.getAllErrors());
 		}
 		final DatasetDTO dataSet = this.middlewareDatasetService.getDataset(datasetId);
-		this.validateDataset(studyId, dataSet, shouldBeSubobservationDataset);
+		this.validateDataset(studyId, dataSet, shouldBeObservationOrSubobservationDataset);
 	}
 
-	private void validateDataset(final Integer studyId, final DatasetDTO dataSet, final Boolean shouldBeSubobservationDataset) {
+	private void validateDataset(
```
clarysabel: @vmaletta same here

src/main/resources/messages_en.properties
```diff
@@ -98,7 +98,7 @@ dataset.type.id.not.exist= {0} is not a valid dataset type
 dataset.does.not.exist=Dataset does not exist
 dataset.required=Dataset is required
 study.required=Study is required
-dataset.type.not.subobservation=Dataset is not a subobservation dataset.
+dataset.type.not.plot.or.subobservation=Dataset is not a plot or subobservation dataset.
```
clarysabel: @vmaletta do not use plot

src/main/java/org/ibp/api/java/impl/middleware/dataset/validator/DatasetValidator.java
clarysabel: @vmaletta When we started the sub-obs dataset development, we added the boolean parameter with the idea of removing it whenever we completed the implementation for the observation scenario. 
Given that (with logical reasons) the QA's are asking for a more specific response code for some scenarios, I think a better idea will be to split the function in 2. One for Datasets and another one for Observations Datasets and remove the boolean parameter. 
It tends to be more confusing to have this boolean 
@nahuel-soldevilla @cuenyad @darla-leafnode for any suggestion

nahuel-soldevilla: Totally agree with removing the flag. As for the message we can keep it generic without the need to have a separate function I think (Maybe renaming the function `validateObservationsDataset`):
```
dataset.type.not.observations=Dataset is not an observations or sub-observations dataset
```
@clarysabel @vmaletta

PR #273, opened 2019-12-20
IBP-3294 Fix Performance of Import Data Discard

abatac:


src/main/java/org/ibp/api/java/impl/middleware/dataset/DatasetServiceImpl.java
```diff
@@ -388,7 +389,11 @@ public void deleteObservation(
 		this.studyValidator.validate(studyId, true);
 		this.datasetValidator.validateDataset(studyId, datasetId, false);
 		this.observationValidator.validateObservation(studyId, datasetId, observationUnitId, observationId, null);
+
+		final Integer observableId = this.middlewareDatasetService.getPhenotype(observationUnitId, observationId).getObservableId();
+
 		this.middlewareDatasetService.deletePhenotype(observationId);
+		this.middlewareDatasetService.updateDependentPhenotypesAsOutOfSync(observableId, observationUnitId);
```
darla-leafnode: We should abstract updating of dependent phenotypes as out of sync from this service. 
That is, it should be part of the Middleware API call.

darla-leafnode:
Closing PR as we removed updating of phenotypes status which was done per phenotype (done one by one before hence the performance issue) when rejecting draft data

PR #278, opened 2020-01-15
IBP-3039 Implement permissions for BMSAPI

darla-leafnode:


darla-leafnode:
@mderamos - please apply code cleanup and formatter on all changed classes

src/main/java/org/ibp/api/java/impl/middleware/crop/CropServiceImpl.java
```diff
@@ -13,6 +14,7 @@
 public class CropServiceImpl implements CropService {
 
 	@Autowired
+	@Lazy
```
clarysabel: @darla-leafnode quick question, why is it @Lazy?

clarysabel:
Hi @darla-leafnode 
I have reviewed the PR and seems fine to me. I left a question just for curiosity. 
Great job to all of you, I have been waiting for this improvement for a while :1st_place_medal:

darla-leafnode:
Thanks @clarysabel for reviewing. 
Regarding use of @Lazy annotation, we used that because were experiencing failing integration tests because of circular dependency issue when autowiring:
`From MiddlewareFactory -> ContextResolver (autowired) -> ProgramService (autowired) -> WorkbenchDataManager(autowired).
` 
`WorkbenchDataManager `could not be autowired from such context because its bean initializer is in `MiddlewareFactory`.

PR #286, opened 2020-02-03
Fix TransactionResource with initial status and type

vmaletta:
Please merge


src/main/java/org/ibp/api/java/impl/middleware/inventory/manager/TransactionServiceImpl.java
```diff
@@ -37,4 +39,14 @@ public Integer saveTransaction(final TransactionDto transactionDto) {
 		this.transactionInputValidator.validate(transactionDto);
 		return this.transactionService.saveTransaction(transactionDto);
 	}
+
+	@Override
+	public List<TransactionStatus>  getAllTransactionStatuses() {
```
clarysabel: Even when statuses is a valid plural, it is uncommon. I think we should use status for plural

src/main/java/org/ibp/api/rest/inventory/manager/TransactionResource.java
```diff
@@ -122,6 +124,9 @@ public long getCount() {
 		if (transactionDto.getLot() == null) {
 			transactionDto.setLot(new ExtendedLotDto());
 		}
+		//FIXME when this resource is completed
+		transactionDto.setTransactionStatus(TransactionStatus.CONFIRMED.getValue());
```
clarysabel: Mention https://ibplatform.atlassian.net/browse/IBP-3455

src/main/java/org/ibp/api/java/impl/middleware/inventory/manager/TransactionServiceImpl.java
vmaletta: Done

src/main/java/org/ibp/api/rest/inventory/manager/TransactionResource.java
vmaletta: Done

PR #291, opened 2020-02-07
IBP-3212 BrAPI v2 study resource get implementation

mderamos:
Kindly review. Thanks!


src/main/java/org/ibp/api/brapi/v2/study/StudyResourceV2Brapi.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.util.Collections;
+import java.util.EnumMap;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@Api(value = "BrAPI v2 Study Services")
+@Controller
+public class StudyResourceV2Brapi {
```
darla-leafnode: Hi @clarysabel - wanted to get your feedback on this new v2 call. Ulysis put it in a new resource class: `StudyResourceV2Brapi ` located under "v2" package. The result is it appears as **study-resource-v-2-brapi** in Swagger. Do you think this is okay versus putting this new service under existing `StudyResourceBrapi `(which is under v1 package) so that all versions of the same service are together?

Copying @nahuel-soldevilla , @abatac too as they might have some feedback too

clarysabel: I will keep both in different packages. If the class is contained in v2 package, I will remove V2 from the class name and add value to the controller so there is no conflict with the beans creation. Having 2 classes with the same name will end having all v1 and v2 services in the same section in Swagger. I will prefer doing this than mixing everything in the v1 packages in our code. 
I wonder if we want to keep same Swagger documentation for v1 and v2? Some projects just split the API documentation per version (we do not have to solve it now, just thinking how we want to handle it?

darla-leafnode: Thanks for your feedback, @clarysabel. I agree as it allows to keep the code more organized.
please take note of the changes versus swagger state when branch testing @christineguiang , @barbacenasmc

darla-leafnode: As for API documentation, I think it's trickier because we have our own BMS versioning.. (v13, v14, v15, etc) and within these versions, there are also different BrAPI versioning. But I agree it would be ideal to somehow split BrAPI v1 and v2 within our own API documentation when uploading to SwaggerHub. I guess we need to spike/explore how to do that, @mcrimi

darla-leafnode:
@mderamos - will merge changes on this PR for now. Please create unit tests tomorrow in another PR

src/main/java/org/ibp/api/brapi/v2/study/StudyResourceV2Brapi.java
clarysabel: @darla-leafnode  Major versions should not be an issue because the specification is incremental

PR #300, opened 2020-02-21
IBP-3491 Exclude servlet-api-2.5 included in Podam

nahuel-soldevilla:


pom.xml
```diff
@@ -198,11 +198,18 @@
 			<version>5.1.0.Final</version>
 		</dependency>
 
+		<!-- Testing: Random Pojo Data Generator -->
```
clarysabel: Hi @abatac  notice that you have included hibernate-validator in Middleware. 
Please remove this artifact from the projects that use Middleware to avoid possible versions conflicts.

PR #304, opened 2020-03-12
IBP-3560 Endpoints should start without '/' 

gelli-leafnode:
Remove '/' from the start of some endpoints.
Testing of brava, endpoints are now called (with missing dependencies).
IBP-3560-EndpointsStartWithoutSlash


src/main/resources/brapi/calls.json
```diff
@@ -14,7 +14,7 @@
           "1.0",
           "1.1",
           "1.2",
-          "1.3",
+          "1.3"
```
darla-leafnode: Are we changing versions too in this PR, @gelli-leafnode ?

gelli-leafnode: 2.0 must retain. Will revert this change. Thanks

PR #317, opened 2020-04-02
Remove Utils.isNullOrEmpty

clarysabel:
IBP-3595

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,6 +39,12 @@
 	<dependencies>
 
 		<dependency>
+    		<groupId>org.springframework</groupId>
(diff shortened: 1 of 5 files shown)


nahuel-soldevilla:
@clarysabel 

As we discussed, we'll wait until we can bump the spring boot version altogether to use the said function, so that we don't have to add spring-core as a separate dependency.
We'll see if we can also improve the pom by grouping all the spring libraries into a [`dependencyManagement`](https://docs.spring.io/spring-boot/docs/current/reference/html/using-spring-boot.html#using-boot-maven-without-a-parent)

PR #319, opened 2020-04-02
IBP-3545 Resolve Brapi Validator issues in supported end points 

gelli-leafnode:
Resolve Brapi validator issues


src/main/resources/brapi/calls.json
```diff
@@ -14,8 +14,7 @@
           "1.0",
           "1.1",
           "1.2",
-          "1.3",
-          "2.0"
```
darla-leafnode, line 18: Why did we remove 2.0 as supported version?

src/main/resources/brapi/calls.json
```diff
@@ -67,6 +66,21 @@
           "1.2"
         ]
       },
+      {
+        "call": "observationunits",
```
darla-leafnode, line 70: Why is there a new endpoint added here?

gelli-leafnode: /observationunits is the new endpoint for phenotypes-search (GET) in [v1.3.](https://brapi.docs.apiary.io/#reference/phenotypes/phenotypes-search). This should be GET not post. The replacement for phenotypes-search (POST) is /search/observationunits. Added this before to pass succeeding observationunit calls. I will remove this endpoint since it's not yet supported.

src/main/resources/brapi/calls.json
gelli-leafnode: Removing 2.0 will resolve the schema mismatch in getCallsResponse.json
image

src/main/java/org/ibp/api/brapi/v1/germplasm/GermplasmResourceBrapi.java
```diff
@@ -125,6 +125,76 @@
 
 	}
 
+	@ApiOperation(value = "Search germplasms", notes = "Replacement for germplasm-search in v1.3")
```
darla-leafnode, line 128: Let's extract changes to this new endpoint to another ticket/PR:
https://ibplatform.atlassian.net/browse/IBP-3624

src/main/resources/brapi/calls.json
darla-leafnode: @nahuel-soldevilla - what do you think about removing 2.0 as supported version in `POST token` call? @gelli-leafnode removed it to resolve schema mismatch error. I guess the BrAPI validator doesn't expect it to be a valid version. Not sure what would be the impact to other consumers if we remove it just to make it pass for the validator.
cc: @mcrimi

nahuel-soldevilla: @darla-leafnode 
I agree to removing it. For v2 we'll need a new endpoint for calls `/v2/serverinfo`

cc @gelli-leafnode @mcrimi

PR #320, opened 2020-04-06
IBP-3185 KSI Field Book - BrAPI Integration

abatac:


src/main/java/org/ibp/api/java/impl/middleware/dataset/validator/ObservationsTableValidator.java
```diff
@@ -83,7 +84,10 @@ private static boolean validateValue(final MeasurementVariable var, final String
 	}
 
 	private static boolean isInvalidDate(final MeasurementVariable var, final String value) {
-		return var.getDataTypeId() != null && var.getDataTypeId() == TermId.DATE_VARIABLE.getId() && !Util.isValidDate(value);
+		// Allow date in yyyy-MM-dd and yyyyDDmm format
+		final Date parsedDate = Util.tryParseDate(value, Util.FRONTEND_DATE_FORMAT);
+		final boolean isDateParseable = parsedDate != null;
+		return var.getDataTypeId() != null && var.getDataTypeId() == TermId.DATE_VARIABLE.getId() && !Util.isValidDate(value) && !isDateParseable;
 	}
```
nahuel-soldevilla, line 91: I wonder if it's possible to change the logic of `Util.isValidDate` instead.
Because if it's not parseable then is must be invalid too.

abatac: Done.

src/main/java/org/ibp/api/brapi/v1/observation/ObservationVariableResult.java
```diff
@@ -7,14 +7,14 @@
...
+	private String studyDbId;
 	private String trialName;
 
-	public int getStudyDbId() {
+	public String getStudyDbId() {
 		return this.studyDbId;
 	}
 
-	public void setStudyDbId(final int studyDbId) {
+	public void setStudyDbId(final int String) {
```
darla-leafnode, line 17: Should be `final String studyDbId`?

PR #323, opened 2020-04-07
IBP-3624-ImplementBrapiGermplasm

gelli-leafnode:
Extracted from IBP-3545; Created new endpoint /germplasm as replacement for germplasm-search for v1.3


darla-leafnode:
Add an entry to `calls.json` for this new service (extract from IBP-3545)

PR #326, opened 2020-04-14
IBP-3248 BrApiMaxPageSize

gelli-leafnode:
Modify for brapi max page size and default page size


src/main/resources/application.properties
```diff
@@ -34,4 +34,6 @@ design.runner=${design.runner}
 design.runner.license.service=${design.runner.license.service}
 
 spring.jackson.mapper.default-view-inclusion=true
-bms.version=${project.version}
\ No newline at end of file
+bms.version=${project.version}
+pagedresult.max.page.size=${pagedresult.max.page.size}
+pagedresult.default.page.size=${pagedresult.default.page.size}
```
darla-leafnode, line 39: Should we remove this:
`pagedresult.default.page.size=${pagedresult.default.page.size} `

gelli-leafnode: Will remove not used property (Defaultpagesize)

src/main/java/org/ibp/api/domain/common/PagedResult.java
```diff
@@ -27,14 +30,14 @@
 	private String sortBy;
 	private String sortOrder;
 
-	public static final int MAX_PAGE_SIZE = 10000;
+	public static int MAX_PAGE_SIZE;
```
darla-leafnode, line 33: I think we can convert MAX_PAGE_SIZE to a non-static local variable such as:
`protected int maxPageSize;`

gelli-leafnode: The instances of PagedResult and BrapiPagedResult (inheriting PagedResult) are being instantiated via new, if we convert the variable MAX_PAGE_SIZE to non-static the value set during spring instantiate will be ignored. Please refer to this [link](https://mkyong.com/spring/spring-inject-a-value-into-static-variables/) for injecting value into static variable

PR #328, opened 2020-04-15
IBP-3514 inventory label printing

nahuel-soldevilla:


src/main/java/org/ibp/api/rest/labelprinting/LotLabelPrinting.java
```diff
@@ -0,0 +1,376 @@
...
+	// Germplasm ids
+	// (Keeping it TERM_ID agnostic for now)
+	private enum GERMPLASM_FIELD {
+		GID(16, "GID"),
+		GROUP_ID(17, "Group ID"),
+		DESIGNATION(18, "Designation"),
+		BREEDING_METHOD(19, "Breeding method"),
+		LOCATION(20, "Location"),
+		CROSS(21, "Cross")
+		// LOT_ID(22) in lot group
```
clarysabel, line 123: do we need to keep this comment?

nahuel-soldevilla: @clarysabel it does't harm and it makes you aware that you cannot use the id 22 at that point and you need to skip it

PR #332, opened 2020-04-21
IBP-3459 API Permissions

mderamos:
Added PreAuthorize Annotations


src/main/java/org/ibp/api/rest/program/ProgramResource.java
```diff
@@ -14,14 +16,14 @@
...
-@RequestMapping("/program")
+@PreAuthorize("hasAnyAuthority('ADMIN','ADMINISTRATION','CROP_MANAGEMENT','SITE_ADMIN')")
 public class ProgramResource {
 
     @Autowired
     private ProgramService programService;
 
-    @RequestMapping(method = RequestMethod.GET)
-    public ResponseEntity<List<ProgramSummary>> listPrograms(@RequestParam final String cropName) {
+    @RequestMapping(value = "/crops/{cropName}/program", method = RequestMethod.GET)
```
darla-leafnode, line 25: Should be /crops/{cropName}/program**s** to follow REST standard

src/main/java/org/ibp/api/rest/program/ProgramResource.java
```diff
@@ -14,14 +16,14 @@
...
     @Autowired
     private ProgramService programService;
 
-    @RequestMapping(method = RequestMethod.GET)
-    public ResponseEntity<List<ProgramSummary>> listPrograms(@RequestParam final String cropName) {
+    @RequestMapping(value = "/crops/{cropName}/programs", method = RequestMethod.GET)
+    public ResponseEntity<List<ProgramSummary>> listPrograms(@PathVariable final String cropName) {
         return new ResponseEntity<>(this.programService.listProgramsByCropName(cropName), HttpStatus.OK);
     }
-}
```
nahuel-soldevilla, line 27: @mderamos @darla-leafnode 

Can we keep both and mark `/program` as `@Deprecated`? For backwards compatibility, maybe there's someone out there using this method :disappointed: 

cc @clarysabel @mcrimi

darla-leafnode: @nahuel-soldevilla - i don't think there's a need to keep it as it's an internal API. It has a counterpart brapi call which is likely the one used: `GET​/{crop}​/brapi​/v1​/programs`

As an aside, in the API permissions work ([IBP-3039](https://ibplatform.atlassian.net/browse/IBP-3039)) for internal services, there were quite a handful of internal services that were renamed to be able to work for crop and program-level permissions checking..
For example all the ontology resources in the old format of "/ontology/{cropname}/xxx" were renamed to "crops/{cropname}/xxx". It's all documented [here](https://ibplatform.atlassian.net/wiki/spaces/DEV/pages/832012689/Implementing+Permissions+in+BMSAPI) so this is actually just a leftover from that analysis and implementation.

nahuel-soldevilla: @darla-leafnode 
Makes sense, and I guess it's not a big deal also, in this case we are just renaming, so if someone is using it they can just adjust accordingly.

PR #333, opened 2020-04-22
Ibp 3667 close lots

clarysabel:
Include:
https://github.com/IntegratedBreedingPlatform/InventoryManager/pull/37
https://github.com/IntegratedBreedingPlatform/Middleware/pull/792


src/main/resources/messages_en.properties
```diff
@@ -420,7 +420,7 @@ lot.withdrawal.input.null=Request Object can not be null
 lot.input.instructions.missing.for.units=Missing instructions for the following unit(s): {0}
 no.lots.selected=List of lots to perform an action can not be empty. Make sure that your searchId returns some lots to work with
 no.transactions.selected=List of transactions to perform an action can not be empty. Make sure that your searchId returns some transactions to work with
-lot.selection.invalid=Either provide searchRequestId or list of lots to operate with
+lot.selection.invalid=Either provide searchId or list of lots to operate with
```
nahuel-soldevilla, line 423: We'll need one for each place we use SearchComposite. Maybe a generic error message would be better. `searchcomposite.invalid=Either select all or choose individual items to operate with` (rewrite message to work either for api usage and for lot UI)

clarysabel: @nahuel-soldevilla this message is for the validation of the comopsite object and it should never be reached when using the api from the inventory manager. So I unified both in the following way:
> search.composite.invalid=Either select a requestId or a set of elements to operate with

clarysabel: @nahuel-soldevilla the one you have suggested could be useful in the frontend when nothing is selected, given that @cuenyad is controlling it and using a different message in the frontend. Opening a discussion!

cuenyad: @nahuel-soldevilla and @clarysabel I think is good idea a generic error message for the API. But for the UI I think would be better a separate messages as is now (Please select at least one lot/Please select at least one transaction) because is related with the interaction with the table instead of the API.

src/main/java/org/ibp/api/java/impl/middleware/inventory/manager/common/SearchRequestDtoResolver.java
```diff
@@ -0,0 +1,45 @@
...
+		final LotsSearchDto searchDTO;
+		if (searchCompositeDto.getSearchRequestId() != null) {
+			searchDTO =
+				(LotsSearchDto) this.searchRequestService.getSearchRequest(searchCompositeDto.getSearchRequestId(), LotsSearchDto.class);
+		} else {
+			searchDTO = new LotsSearchDto();
+			searchDTO.setLotIds(new ArrayList<>(searchCompositeDto.getItemIds()));
+		}
+		return searchDTO;
+	}
```
nahuel-soldevilla, line 30: I wonder if something like this is possible. But we'll the dtos to implement a `setItemIds` 
```java
	public T getLotsSearchDto(final SearchCompositeDto searchCompositeDto) {

		final T searchDTO;
		if (searchCompositeDto.getSearchRequestId() != null) {
			searchDTO = (T) this.searchRequestService.getSearchRequest(searchCompositeDto.getSearchRequestId(), LotsSearchDto.class);
		} else {
			searchDTO = new T();
			searchDTO.setItemIds(new ArrayList<>(searchCompositeDto.getItemIds()));
		}
		return searchDTO;
	}
```

PR #335, opened 2020-04-27
IBP-3670 Handle NA (Not Available) value in PUT Observations

abatac:


src/main/java/org/ibp/api/java/impl/middleware/dataset/validator/ObservationsTableValidator.java
```diff
@@ -70,7 +71,7 @@ public void validateObservationsValuesDataTypes(final Table<String, String, Stri
 	}
 
 	private static boolean validateValue(final MeasurementVariable var, final String value, final BindingResult errors) {
-		if (StringUtils.isBlank(value)) {
+		if (StringUtils.isBlank(value) || MISSING_VALUE.equals(value)) {
```
darla-leafnode, line 74: Placing this here could allow "missing" value for date variables

PR #337, opened 2020-04-27
Token for BrAPI v2

clarysabel:
Issue: IBP-3699 / IBP-3566


clarysabel:
Quick question, given that we are supporting token for v2, does the calls.json file needs to be modified? @nahuel-soldevilla

PR #340, opened 2020-04-30
IBP-3691 Non-owner of a study should not be able to open/sync data of a locked study

abatac:


src/main/java/org/ibp/api/java/impl/middleware/dataset/validator/StudyValidator.java
```diff
@@ -50,14 +50,8 @@ public void validate(final Integer studyId, final Boolean shouldBeUnlocked) {
...
-		final WorkbenchUser loggedInUser = this.securityService.getCurrentlyLoggedInUser();
-
-		if (shouldBeUnlocked
-			&& study.isLocked()
-			&& !study.getCreatedBy().equals(loggedInUser.getUserid().toString())
-			&& !loggedInUser.isSuperAdmin()) {
-			this.errors.reject("study.is.locked", "");
-			throw new ForbiddenException(this.errors.getAllErrors().get(0));
+		if (shouldBeUnlocked) {
+			this.checkIfStudyIsUnlocked(study);
```
darla-leafnode, line 54: checkIfStudyIsUnlocked is inaccurate because the user can still edit a locked study if he is the owner or is a superadmin

src/main/java/org/ibp/api/brapi/v1/study/StudyResourceBrapi.java
```diff
@@ -230,6 +234,10 @@ private StudyObservations getStudyObservations(final int studyDbId) throws Excep
 	@JsonView(BrapiView.BrapiV1_3.class)
 	public ResponseEntity<StudyDetails> getStudyDetails(@PathVariable final String crop, @PathVariable final Integer studyDbId) {
 
+		// Study should be unlocked before retrieving the study details
+		final Integer studyId = this.studyDataManager.getProjectIdByStudyDbId(studyDbId);
```
darla-leafnode, line 238: Pkease remove the study status checking on these GET functions

PR #346, opened 2020-06-02
Using InventoryCommonValidator for notes length and units

clarysabel:
IBP-3779


clarysabel:
Right, I could have removed some tests because they are covered in InventoryCommonValidatorTest, so calling the real method was not needed. I just wanted to keep the same set of testing we had.

PR #348, opened 2020-06-05
IBP-3760 enable the use lot UUID to external clients

cuenyad:
- Replacing the using lotUUID instead of LotID in SearchCompositeDto.
- Replacing the using lotId to validate the lot and use the lotUUID instead.
- Modifying Create Lot Resource to return the LotUUID generated instead of LotID.

Includes:
https://github.com/IntegratedBreedingPlatform/InventoryManager/pull/49
https://github.com/IntegratedBreedingPlatform/Middleware/pull/814

Please, Review


src/main/java/org/ibp/api/rest/inventory/manager/LotResource.java
```diff
@@ -278,5 +280,27 @@ public long getCount() {
...
 
+	@ApiOperation(value = "It will retrieve a lot", notes = "It will retrieve lot by lotUUID")
+	@RequestMapping(value = "/crops/{cropName}/lots/{lotUUID}", method = RequestMethod.GET)
+	@PreAuthorize(HAS_MANAGE_LOTS + " or hasAnyAuthority('VIEW_LOTS')")
+	@ResponseBody
+	@JsonView(InventoryView.LotView.class)
+	public ResponseEntity<ExtendedLotDto> getLot(@PathVariable final String cropName, @PathVariable final String lotUUID) {
+
+		final LotsSearchDto searchDTO = new LotsSearchDto();
+		searchDTO.setLotUUIDs(Arrays.asList(lotUUID));
```
clarysabel, line 291: @cuenyad  It is ok, but when creating a list for a single element I prefer to use Collections.singletonList
https://stackoverflow.com/questions/26027396/arrays-aslist-vs-collections-singletonlist

PR #352, opened 2020-06-16
IBP-3388 Optimize Execute Calculated Variable

mderamos:


src/main/java/org/ibp/api/java/impl/middleware/derived/DerivedVariableServiceImpl.java
```diff
@@ -215,6 +217,9 @@ public DerivedVariableServiceImpl() {
 
 		}
 
+		// Update the dependent phenotypes as out of sync in batch
+		this.middlewareDatasetService.updateDependentPhenotypesAsOutOfSync(targetMeasurementVariable.getTermId(), observationUnitIds);
```
abatac, line 221: Hi @mderamos modify execute() method's unit test to verify middlewareDatasetService.updateDependentPhenotypesAsOutOfSync is called.

PR #357, opened 2020-07-01
IBP-3827 Throw error 403 for unauthorized access

mderamos:


mderamos:
Hi @nahuel-soldevilla, it does not catch authentication errors. But I think you're right. I should throw org.springframework.security.access.AccessDeniedException.

PR #364, opened 2020-07-10
IBP-3830 IBP-3831 BrAPI Get Lots and Get Transactions

darla-leafnode:


src/main/java/org/ibp/api/brapi/v2/inventory/LotDetails.java
```diff
@@ -0,0 +1,162 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+@AutoProperty
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class LotDetails {
+
+    private Double amount;
+
+    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss.SSSXXX")
```
abatac, line 18: Date format should be yyyy-MM-dd only. Based on this spec: https://github.com/plantbreeding/API/blob/master/Specification/GeneralInfo/Date_Time_Encoding.md

src/main/java/org/ibp/api/brapi/v2/inventory/TransactionDto.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.Map;
+
+@AutoProperty
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class TransactionDto {
+
+	private Map<String, Object> additionalInfo;
+	private Double amount;
+	private String transactionDescription;
+	@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyyMMdd")
```
abatac, line 17: Date format should be yyyy-MM-ddThh:mm:ssz. Based on this spec: https://github.com/plantbreeding/API/blob/master/Specification/GeneralInfo/Date_Time_Encoding.md

clarysabel:
Hi @darla-leafnode  and @abatac 
In general looks fine, but you still need to add the 2 new services to the calls call. It was only added for testing purposes but when calling calls the 2 new services are lot listed. 
Once this is addressed, you can merge it!

PR #373, opened 2020-07-30
IBP 3867 Create Lots from Crosses and selections Tab

abatac:


src/main/java/org/ibp/api/java/impl/middleware/inventory/manager/TransactionServiceImpl.java
```diff
@@ -156,6 +160,9 @@ public void updatePendingTransactions(final List<TransactionUpdateRequestDto> tr
 	public void saveDeposits(final LotDepositRequestDto lotDepositRequestDto, final TransactionStatus transactionStatus) {
 		final WorkbenchUser user = this.securityService.getCurrentlyLoggedInUser();
 
+		if (lotDepositRequestDto.getSourceStudy() != null) {
+			this.studyValidator.validate(lotDepositRequestDto.getSourceStudy(), false);
```
darla-leafnode, line 164: Since we're making experiment transaction record for the study, I think study should be unlocked if study is specified.

src/main/java/org/ibp/api/java/impl/middleware/inventory/manager/TransactionServiceImpl.java
```diff
@@ -156,6 +160,9 @@ public void updatePendingTransactions(final List<TransactionUpdateRequestDto> tr
 	public void saveDeposits(final LotDepositRequestDto lotDepositRequestDto, final TransactionStatus transactionStatus) {
```
darla-leafnode, line 160: Can we make a unit test for saveDeposits, considering the test case that source study may or may not have been specified?

src/main/java/org/ibp/api/java/impl/middleware/inventory/manager/TransactionServiceImpl.java
abatac: Done

src/main/java/org/ibp/api/java/impl/middleware/inventory/manager/TransactionServiceImpl.java
abatac: Done

PR #396, opened 2020-08-27
IBP-4007 fields with alias names does not display the values on export labels

cuenyad:
Please, review.


clarysabel:
Agree. We could include it in 16.2

PR #408, opened 2020-10-07
Ibp 4053 inventory manager split lot

lik-83:


src/main/java/org/ibp/api/java/impl/middleware/inventory/manager/LotServiceImpl.java
```diff
@@ -217,4 +232,54 @@ public void mergeLots(final String keepLotUUID, final LotsSearchDto lotsSearchDt
...
+		//Creates the new lot
+		final LotSplitRequestDto.NewLotSplitDto newLot = lotSplitRequestDto.getNewLot();
+		final LotGeneratorInputDto lotGeneratorInputDto = new LotGeneratorInputDto();
+		lotGeneratorInputDto.setGid(splitLotDto.getGid());
+		lotGeneratorInputDto.setUnitId(splitLotDto.getUnitId());
+		lotGeneratorInputDto.setLocationId(newLot.getLocationId());
+		lotGeneratorInputDto.setNotes(newLot.getNotes());
+		lotGeneratorInputDto.setGenerateStock(newLot.getGenerateStock());
+		lotGeneratorInputDto.setStockPrefix(newLot.getStockPrefix());
+		String newLotUUID = this.saveLot(programUUID, lotGeneratorInputDto);
```
clarysabel, line 259: Instead of calling this.saveLot I will prefer to use lotService.saveLot. Extracting the stockId generation to a new reusable function. This will allow the validations to run first of any save in the database and avoid calling unnecessary validations like unit Id

src/main/java/org/ibp/api/java/impl/middleware/inventory/manager/LotServiceImpl.java
```diff
@@ -217,4 +232,54 @@ public void mergeLots(final String keepLotUUID, final LotsSearchDto lotsSearchDt
...
+		final List<ExtendedLotDto> extendedLotDtos = this.searchLots(lotsSearchDto, null);
+		this.extendedLotListValidator.validateAllProvidedLotUUIDsExist(extendedLotDtos, new HashSet<>(lotUUIDs));
+
+		ExtendedLotDto splitLotDto = extendedLotDtos.get(0);
+		LotSplitRequestDto.InitialLotDepositDto initialDeposit = lotSplitRequestDto.getInitialDeposit();
+		this.lotSplitValidator.validateSplitLot(splitLotDto, initialDeposit.getAmount());
+
+		//Creates the new lot
+		final LotSplitRequestDto.NewLotSplitDto newLot = lotSplitRequestDto.getNewLot();
+		final LotGeneratorInputDto lotGeneratorInputDto = new LotGeneratorInputDto();
```
clarysabel, line 252: Create a constructor

src/main/java/org/ibp/api/java/impl/middleware/inventory/manager/LotServiceImpl.java
```diff
@@ -217,4 +232,54 @@ public void mergeLots(final String keepLotUUID, final LotsSearchDto lotsSearchDt
...
+		String newLotUUID = this.saveLot(programUUID, lotGeneratorInputDto);
+
+		//Create an adjustment transaction for the split lot
+		LotAdjustmentRequestDto lotAdjustmentRequestDto = new LotAdjustmentRequestDto();
+		lotAdjustmentRequestDto.setBalance(splitLotDto.getActualBalance() - initialDeposit.getAmount());
+
+		SearchCompositeDto<Integer, String> balanceAdjustmentSearchCompositeDto = new SearchCompositeDto<>();
+		balanceAdjustmentSearchCompositeDto.setItemIds(Arrays.asList(splitLotDto.getLotUUID()).stream().collect(Collectors.toSet()));
+		lotAdjustmentRequestDto.setSelectedLots(balanceAdjustmentSearchCompositeDto);
+		this.transactionService.saveLotBalanceAdjustment(lotAdjustmentRequestDto);
```
clarysabel, line 268: same here, I think calling transactionService.saveAdjustmentTransactions directly will work better and all validations rules are located before any call to middleware

src/main/java/org/ibp/api/java/impl/middleware/inventory/manager/LotServiceImpl.java
```diff
@@ -217,4 +232,54 @@ public void mergeLots(final String keepLotUUID, final LotsSearchDto lotsSearchDt
...
+		final Map<String, Double> depositsPerUnit = new HashMap() {{
+			put(splitLotDto.getUnitName(), initialDeposit.getAmount());
+		}};
+		lotDepositRequestDto.setDepositsPerUnit(depositsPerUnit);
+		lotDepositRequestDto.setNotes(initialDeposit.getNotes());
+
+		SearchCompositeDto<Integer, String> saveDepositSearchCompositeDto = new SearchCompositeDto<>();
+		saveDepositSearchCompositeDto.setItemIds(Arrays.asList(newLotUUID).stream().collect(Collectors.toSet()));
+		lotDepositRequestDto.setSelectedLots(saveDepositSearchCompositeDto);
+		this.transactionService.saveDeposits(lotDepositRequestDto, TransactionStatus.CONFIRMED, TransactionSourceType.SPLIT_LOT,
```
clarysabel, line 281: same here!

src/main/java/org/ibp/api/java/impl/middleware/inventory/manager/TransactionServiceImpl.java
```diff
@@ -167,7 +168,8 @@ public void updatePendingTransactions(final List<TransactionUpdateRequestDto> tr
 	}
 
 	@Override
-	public void saveDeposits(final LotDepositRequestDto lotDepositRequestDto, final TransactionStatus transactionStatus) {
+	public void saveDeposits(final LotDepositRequestDto lotDepositRequestDto, final TransactionStatus transactionStatus,
```
clarysabel, line 171: If we call MW directly probably this signature does not need to change?

clarysabel (changes requested):
Hi @lik-83 I have left some comments in this PR

clarysabel:
I think it is fine, my only suggestion is that we should provide some basic functions in lots and transactions like: create a new transaction or a new lot that does not need a complex object. But it can be done later

nahuel-soldevilla:
> I think it is fine, my only suggestion is that we should provide some basic functions in lots and transactions like: create a new transaction or a new lot that does not need a complex object. But it can be done later

That's fine, thought some times we only use these constructors for testing, in some places we use some of the params, and in other places we use others. Some times we end up with classes with many constructors (like org.generationcp.middleware.domain.etl.MeasurementData) and then it's difficult to know which one to use. But if the constructor is needed to put a constraint in the way some object should be created, then I think it makes sense.

@clarysabel

PR #411, opened 2020-10-14
IBP-4029 convert germplasm tab to angular table

mderamos:


src/main/java/org/ibp/api/rest/germplasm/EntryTypeResource.java
```diff
@@ -0,0 +1,67 @@
...
+	@ResponseBody
+	public ResponseEntity<List<Enumeration>> getStudyEntryTypes(final @PathVariable String cropname,
+		@PathVariable final String programUUID) {
+		final List<Enumeration> entryTypes =
+			this.entryTypeService.getEntryTypes(programUUID);
+		return new ResponseEntity<>(entryTypes, HttpStatus.OK);
+	}
+
+	@ApiOperation(value = "Add Entry Type",
+		notes = "Add or update Study Entry Type")
```
abatac, line 39: Change to just "Add Study Entry Type"

src/main/java/org/ibp/api/rest/germplasm/EntryTypeResource.java
```diff
@@ -0,0 +1,67 @@
...
+	@RequestMapping(value = "/{cropname}/programs/{programUUID}/entry-types", method = RequestMethod.POST)
+	@ResponseBody
+	public ResponseEntity addEntryType(final @PathVariable String cropname,
+		@PathVariable final String programUUID,	@RequestBody Enumeration entryType) {
+		this.entryTypeService.addEntryType(programUUID, entryType);
+		return new ResponseEntity<>(HttpStatus.NO_CONTENT);
+	}
+
+	@ApiOperation(value = "Add or update Study Entry Type",
+		notes = "Add or update Study Entry Type")
```
abatac, line 49: Change to just "Update Study Entry Type"

src/main/java/org/ibp/api/java/impl/middleware/study/StudyEntryServiceImpl.java
```diff
@@ -140,6 +147,26 @@ public long countAllStudyEntries(final Integer studyId) {
...
+
+	@Override
+	public long countAllCheckTestEntries(final Integer studyId, final String programUuid, final Boolean checkOnly) {
+		final List<Enumeration> entryTypes = this.entryTypeService.getEntryTypes(programUuid);
+		if(checkOnly) {
+			return this.middlewareStudyEntryService.countStudyGermplasmByEntryTypeIds(studyId,
+				Collections.singletonList(String.valueOf(SystemDefinedEntryType.CHECK_ENTRY.getEntryTypeCategoricalId())));
+		} else {
+			final List<String> checkEntryTypeIds = entryTypes.stream()
+				.filter(enttryType -> enttryType.getId() != SystemDefinedEntryType.TEST_ENTRY.getEntryTypeCategoricalId())
```
clarysabel, line 164: typo enttryType @mderamos

src/main/java/org/ibp/api/rest/study/StudyEntryResource.java
```diff
@@ -143,4 +145,23 @@ public long getCount() {
 
 		return new ResponseEntity<>(entryDescriptors, HttpStatus.OK);
 	}
+
+	@ApiOperation(value = "Get Study Test Entries count",
+		notes = "Get Study Test Entries count")
+	@PreAuthorize("hasAnyAuthority('ADMIN','STUDIES','MANAGE_STUDIES')")
+	@RequestMapping(value = "/{crop}/programs/{programUUID}/studies/{studyId}/entries/count-test-entries", method = RequestMethod.GET)
```
clarysabel, line 152: It is not clear to me why we need these 2 services. Should not be better just to add this information in the header response of the GET entries? it is metadata of the GET service right? 
If that is not possible I think a better option is to create /entries/metadata service and return all the counts inside, will be more REST and BMSAPI compliant 
@darla-leafnode @mderamos

PR #412, opened 2020-10-14
IBP-4071 IBP-3939 Germplasm Search

abatac:


src/main/java/org/ibp/api/rest/breedingmethod/BreedingMethodResource.java
```diff
@@ -0,0 +1,70 @@
...
+
+	@ApiOperation(value = "List breeding method types")
+	@RequestMapping(value = "/crops/{cropName}/breedingmethod-types", method = RequestMethod.GET)
+	@ResponseBody
+	public ResponseEntity<List<MethodType>> getBreedingMethodTypes(
+		@PathVariable final String cropName,
+		@RequestParam(required = false) final String programUUID
+	) {
+		return new ResponseEntity<>(MethodType.getAll(), HttpStatus.OK);
+	}
```
nahuel-soldevilla, line 49: @clarysabel What do you think of that idea I told you once, about having XXXResourceGroup classes, to group resources that are closely related? So that we don't have to move them to different files? For this case I could rename the class to `BreedingMethodResourceGroup`

clarysabel: I think it is fine but should not be better to name it XXXGroupResource? so we keep Resource at the end?

PR #417, opened 2020-10-26
IBP-3944 save germplasm list

nahuel-soldevilla:
- germplasm list types resource
- create germplasm list resource
  - accepts optional fields to support germplasm import later
- new `ApiValidationException` for single validations


src/main/java/org/ibp/api/rest/germplasmlist/GermplasmListResourceGroup.java
```diff
@@ -37,4 +41,29 @@
 		final List<TreeNode> children = this.germplamListService.getGermplasmListChildrenNodes(crop, programUUID, parentFolderId, onlyFolders);
 		return new ResponseEntity<>(children, HttpStatus.OK);
 	}
+
+	@ApiOperation(value = "Create a new Germplasm list")
+	@RequestMapping(value = "/crops/{crop}/germplasm-lists", method = RequestMethod.POST)
+	@PreAuthorize("hasAnyAuthority('ADMIN', 'CROP_MANAGEMENT', 'GERMPLASM', 'MANAGE_GERMPLASM')")
+	@ResponseBody
+	public ResponseEntity<GermplasmListGeneratorDTO> create(
```
clarysabel, line 49: Is there any reason why the service is responding with GermplasmListGeneratorDTO instead of GermplasmList? would it make more sense?

src/main/java/org/ibp/api/java/impl/middleware/germplasm/GermplamListServiceImpl.java
```diff
@@ -123,6 +158,157 @@ public GermplasmList getGermplasmList(final Integer germplasmListId) {
...
+			}
+
+			if (isBlank(entry.getSeedSource())) {
+				entry.setSeedSource(GermplasmListDataDAO.SOURCE_UNKNOWN);
+				hasSeedSourceEmpty = true;
+			} else {
+				hasSeedSource = true;
+			}
+
+			if (isBlank(entry.getDesignation())) {
```
clarysabel, line 265: Same question here, I think we should not even consider the Designation to be set by the users if the service is used by an external client

clarysabel (commented):
Leaving some comments

src/main/java/org/ibp/api/java/impl/middleware/germplasm/GermplamListServiceImpl.java
nahuel-soldevilla: I'll remove it

src/main/java/org/ibp/api/rest/germplasmlist/GermplasmListResourceGroup.java
nahuel-soldevilla: Maybe, though in that case perhaps the parameter should also be a GermplasmListDTO, since it's a POST /germplasm-list.
Another option is POST /germplasm-lists/generation, in which case it makes more sense to respond with the generatorDTO.
In any case, it doesn't make much of a difference, it's just a way to answer to the client with the same object with the id filled in (and the entryNo if auto-generated)

@clarysabel

PR #425, opened 2020-11-05
IBP-4084 Import Germplasm Updates

abatac:


src/main/java/org/ibp/api/java/impl/middleware/common/validator/GermplasmUpdateValidator.java
```diff
@@ -0,0 +1,152 @@
...
+
+	}
+
+	public void validateLocationAbbreviation(final BindingResult errors, final List<GermplasmUpdateDTO> germplasmUpdateDTOList) {
+
+		final Set<String> locationAbbreviationSet =
+			germplasmUpdateDTOList.stream().filter(dto -> StringUtils.isNotEmpty(dto.getLocationAbbreviation()))
+				.map(dto -> dto.getLocationAbbreviation()).collect(Collectors.toSet());
+		final List<String> abbreviations =
+			this.locationDataManager.getLocationsByAbbreviation(locationAbbreviationSet).stream().map(loc -> loc.getLabbr()).collect(
```
clarysabel, line 113: Please consider that we need to use the programUUID to validate the locations, because the list changes depending the program and user that does not have access to an specific program location should not be using that code. The same with breeding methods. @abatac

abatac: Noted on this. I'll update this once your branch is merged.

abatac: Done.

src/main/java/org/ibp/api/rest/germplasm/GermplasmResource.java
```diff
@@ -182,4 +183,16 @@ public long getFilteredCount() {
 		return new ResponseEntity<>(this.germplasmService.importGermplasm(cropName, programUUID, germplasmImportRequestDto), HttpStatus.OK);
 	}
 
+	@ApiOperation(value = "Import germplasm updates")
+	@RequestMapping(value = "/crops/{cropName}/germplasm", method = RequestMethod.PATCH)
+	@ResponseBody
+	public ResponseEntity<Void> importGermplasmUpdates(@PathVariable final String cropName,
+		@RequestParam(required = false) final String programUUID,
+		@RequestBody final List<GermplasmUpdateDTO> germplasmUpdateDTOList) {
```
clarysabel, line 191: I suggest to rename this germplasmUpdateDTOList to germplasmList (or maybe there is anything in Swagger that allows us to show a different variable name) Not very important though

src/main/java/org/ibp/api/java/impl/middleware/common/validator/GermplasmUpdateValidator.java
```diff
@@ -0,0 +1,171 @@
...
+	@Autowired
+	private BreedingMethodService breedingMethodService;
+
+	public void validateEmptyList(final BindingResult errors, final List<GermplasmUpdateDTO> germplasmUpdateDTOList) {
+		if (germplasmUpdateDTOList == null || germplasmUpdateDTOList.isEmpty()) {
+			errors.reject("germplasm.update.empty.list", "");
+		}
+	}
+
+	public void validateCodes(final BindingResult errors, final List<GermplasmUpdateDTO> germplasmUpdateDTOList) {
```
clarysabel, line 52: Maybe I am missing something with this function. How an user get ambiguosCodes? 
You will have a file and how given the columns you will build the names and attributes map?

src/main/java/org/ibp/api/java/impl/middleware/common/validator/GermplasmUpdateValidator.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+@Component
+public class GermplasmUpdateValidator {
```
clarysabel, line 29: General comment to standarize our code
I am considering the keys of the map case insensitive. For example, if an user use DRVNAME or drvname to me is the same. 
But given that when we get values for the map the comparisson is made by equals I am using toUpperCase in many places of the code 
Not sure if you handle this case and what is the expected result if an user enter twice the same attribute or name? Example "drvname", "DRVNAME"? Trying to update the same name type twice ? 
In my case I just reject the request because the key is duplicated.
The same with the preferredName, if the keyName is upper case and preferred name is lower case it is still valid to me. I Can see this onconsistency:

This is valid: 
```
    "names": {
      "LNAME": "NewLName",
      "accname": "New preferred name"
    },
    "preferredName": "ACCNAME",
```

But this is not: 
```
    "names": {
      "LNAME": "NewLName",
      "accname": "New preferred name"
    },
    "preferredName": "accname",
```

clarysabel (commented):
Leaving some comments

src/main/java/org/ibp/api/java/impl/middleware/common/validator/GermplasmUpdateValidator.java
abatac: This is specifically for the case where an FCODE is defined as Name and Attribute in the udflds. Yeah I think we can move this checking in the client when building the map for names and attributes.

src/main/java/org/ibp/api/java/impl/middleware/common/validator/GermplasmUpdateValidator.java
abatac: I'll update the code make the fcode case insensitive as part of IBP-4154.

PR #430, opened 2020-11-10
IBP-4171 add validation on generate study entries list

mderamos:


src/main/java/org/ibp/api/java/impl/middleware/study/validator/StudyEntryValidator.java
```diff
@@ -83,6 +83,20 @@ public void validate(final Integer studyId, final Integer entryId, final Integer
 
 	}
 
+	public void validateStudyAlreadyHasStudyEntries(final Integer studyId) {
+
+		this.errors = new MapBindingResult(new HashMap<String, String>(), Integer.class.getName());
+
+		if (this.middlewareStudyEntryService.countStudyEntries(studyId) > 0) {
+			errors.reject("study.has.existing.study.entries");
```
darla-leafnode: You can throw exception inside this if clause already since you are just doing 1 validation

darla-leafnode (approved):

PR #431, opened 2020-11-11
IBP-4164 fix elo ordering

mderamos:


src/main/java/org/ibp/api/java/impl/middleware/design/type/EntryListOrderDesignTypeServiceImpl.java
```diff
@@ -115,6 +115,9 @@ ObservationUnitRow createObservationUnitRow(final int instanceNumber, final Stud
 	private void loadChecksAndTestEntries(final List<StudyEntryDto> studyEntryDtoList, final List<StudyEntryDto> checkList,
 		final List<StudyEntryDto> testEntryList) {
 
+		// Sort Study Entry List by entry number
+		studyEntryDtoList.sort(Comparator.comparing(StudyEntryDto::getEntryNumber));
```
darla-leafnode, line 119: I don't understand why we need to perform sorting here. I expect that `StudyEntryServiceImpl#getStudyEntries` should return the entries sorted by entry number and in fact I see this code that's supposed to do that:
`// Return by ascending order of entry number. We need to perform cast first on uniquename since it's stored as string
		return this.getStudyEntries(studyId, null, new PageRequest(0, Integer.MAX_VALUE,
				new Sort(Sort.Direction.ASC, String.valueOf(TermId.ENTRY_NO.getId()))));`
Is it not doing that?

darla-leafnode:
As discussed, the fix should be in Middleware service. The entries should be ordered there.

PR #433, opened 2020-11-18
IBP-4043 Mark as Deprecated.

cuenyad:
Deprecate Resource using instead "/token"

Include:
https://github.com/IntegratedBreedingPlatform/Workbench/pull/388

Please, review
Regards, Diego

### DIFF
--- a/src/main/java/org/ibp/api/security/xauth/UserXAuthTokenController.java
+++ b/src/main/java/org/ibp/api/security/xauth/UserXAuthTokenController.java
@@ -1,6 +1,7 @@
 
 package org.ibp.api.security.xauth;
 
+import io.swagger.annotations.ApiOperation;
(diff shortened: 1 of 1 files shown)


nahuel-soldevilla (approved):

src/main/java/org/ibp/api/security/xauth/UserXAuthTokenController.java
```diff
@@ -31,6 +31,7 @@
 	@Autowired
 	private UserDetailsService userDetailsService;
 
+	@Deprecated
 	@RequestMapping(value = "/authenticate", method = RequestMethod.POST)
```
nahuel-soldevilla, line 35: We can leave a note:
```java
    @ApiOperation(notes = "<b>DEPRECATED:</b> use /token ")
```

PR #436, opened 2020-11-27
IBP-4150 Germplasm Search improvements

darla-leafnode:


src/main/java/org/ibp/api/rest/germplasm/GermplasmResource.java
```diff
@@ -142,6 +142,15 @@ public long getFilteredCount() {
 		return new ResponseEntity<>(this.germplasmService.filterGermplasmNameTypes(codes), HttpStatus.OK);
 	}
 
+	@ApiOperation(value = "Search germplasm name types")
+	@RequestMapping(value = "/crops/{cropName}/germplasm/name-types/search", method = RequestMethod.GET)
+	@ResponseBody
+	public ResponseEntity<List<GermplasmNameTypeDTO>> searchNameTypes(@PathVariable final String cropName,
+															   @RequestParam(required = false) final String programUUID,
```
abatac, line 149: Is programUUID needed in this method?

darla-leafnode: It is an optional request parameter, as the other endpoints in this resource. I believe it's mainly for retrieving program-level permissions

PR #442, opened 2020-12-02
Ibp 4184 add germplasm entries to list

lik-83:


src/main/java/org/ibp/api/rest/germplasmlist/GermplasmListResourceGroup.java
```diff
@@ -64,6 +66,19 @@
 		return new ResponseEntity<>(germplasmListTypes, HttpStatus.OK);
 	}
 
+	@ApiOperation(value = "Add germplasm entries to an existing list")
+	@RequestMapping(value = "/crops/{crop}/germplasm-list/{germplasmListId}/entries", method = RequestMethod.POST)
```
clarysabel, line 70: rename germplasm-list to germplasm-lists @lik-83

src/main/resources/messages_en.properties
```diff
@@ -462,6 +462,9 @@ list.delete.not.folder=Item is not a folder
 list.delete.root.folder.not.allowed=Root folder can not be deleted
 list.delete.not.owner=You cannot delete a list that you do not own
 list.delete.folder.has.child=Unable to delete folder because it has items.
+list.ids.invalid=The following list id(s): {0} were not found in the system
+list.invalid=ListId is not a list
+list.locked=List is locked
```
clarysabel, line 467: ListId is not a list can be replaced by: We have not found the list id {0} in the system
@lik-83

PR #444, opened 2020-12-04
IBP-3919 Register 2.0 support for GET /studies in calls.json

darla-leafnode:
Hi @clarysabel - kindly merge.


src/main/resources/brapi/calls.json
```diff
@@ -155,6 +156,7 @@
         "dataTypes": [
           "application/json"
         ],
+        ],
```
clarysabel, line 159: @darla-leafnode I think this extra line is not ok, could you please check?

PR #446, opened 2020-12-09
IBP-4003-SelectAllPages

gelli-leafnode:
Used GermplasmStudySourceService to fetch GID related to study 
Added studyId in LotsSearchDto for api POST "/crops/{cropName}/lots/search"
Remove studyId value, replaced by gids


src/main/java/org/ibp/api/rest/inventory/manager/LotResource.java
```diff
@@ -125,6 +131,17 @@
 		@PathVariable final String cropName,
 		@RequestParam(required = false) final String programUUID,
 		@RequestBody final LotsSearchDto lotsSearchDto) {
+
+		if (lotsSearchDto.getStudyId() != null) {
```
abatac, line 135: Put a comment on this code to explain the reason why we need to get all gids from studyId.

src/main/java/org/ibp/api/rest/inventory/manager/LotResource.java
```diff
@@ -125,6 +131,17 @@
...
+			final GermplasmStudySourceSearchRequest studySourceSearchRequest = new GermplasmStudySourceSearchRequest();
+			studySourceSearchRequest.setStudyId(Integer.valueOf(lotsSearchDto.getStudyId()));
+			final List<GermplasmStudySourceDto> gids = this.germplasmStudySourceService.getGermplasmStudySources(studySourceSearchRequest,null);
+			if (gids != null) {
+				lotsSearchDto.setGids(gids.stream().map(GermplasmStudySourceDto::getGid).collect(Collectors.toList()));
+			}
+			// Removing studyId value to match gidsearchdto
+			lotsSearchDto.setStudyId(null);
+		}
```
nahuel-soldevilla, line 144: Hi @gelli-leafnode ,

I have some concerns about this.
First: This only works because the gids field is available with the same name both in **GidSearchDto and LotsSearchDto**, right?
I'm guessing the deserialization process doesn't care as long long as it has the same fields? This approach looks exceedingly **fragile** if that's the case.

Second, supposing that we need a `studyId` filter in searchLots (which would be like the the union of the current harvesting/planting stutydIds filters), I think it'd be probably better to implement it directly as a filter in the query, without going through germplasmStudySource first.

I agree that the problem we are solving is not trivial though.
As we move to complete frontend/backend separation, with mandatory pagination as a feature, we encounter these issues were the frontend doesn't have all the necessary info, like the list of gids. In the germplasm import we faced a similar problem, and since saving a searchId was not an option in that case, we end up collecting all records with multiple ajax calls: https://github.com/IntegratedBreedingPlatform/Workbench/commit/10617625e41ff09c5a8d3fe93c2c4d5bea68a99f

For this case, searchId is an option but we need to translate between germplasmStudySource search to list of gids, as you did here.

What other options do we have?

We can move this translation logic from the search lots to the createLots method. E.g we add a studyId parameter to LotGeneratorBatchRequestDto, and if it comes with it (always if it comes from manage studies) and the searchComposite is null (no individual list of gids set), we resolve from a GermplasmStudySourceSearchRequest.studyId search (or even better: a new studyId filter in lot search, as mentioned above).
Would that work?
It'd be something like this
```javascript
// studies
$scope.openLotCreationModal = function () {
    // ... validate selections first
    
    let promise;
    if ($scope.isAllPagesSelected) {
       promise = Promise.resolve(null);
    } else {
       promise = lotService.saveSearchRequest({gids: Object.keys($scope.selectedItems))
    }
    promise.then((searchDto) => {
       // ... open popup
```

```java
// LotServiceImpl#createLots (after saving from lot-creation component)
if (lotGeneratorBatchRequestDto.getStudyId()) {
    final GermplasmStudySourceSearchRequest studySourceSearchRequest = new GermplasmStudySourceSearchRequest();
    studySourceSearchRequest.setStudyId(Integer.valueOf(lotsSearchDto.getStudyId()));
    ...
} else {
    this.searchRequestDtoResolver.resolveGidSearchDto(searchComposite)
...
```
----
Finally, I'm not sure if you already started working on it, but the **keyboard shortcuts part (shift-click)** is more like a feature request to me now that we switched implementations. I know that it was possible before in the update inventory, but I'm not sure we should address it for bug fixing (unless a simple solution is available). Moreover, we still need to think on how to address it in new angular > 2 modules.

----
@abatac sorry I deleted your comment by mistake and there's no way to undo it 

>  		@PathVariable final String cropName,
>  		@RequestParam(required = false) final String programUUID,
>  		@RequestBody final LotsSearchDto lotsSearchDto) {
> 
> 		if (lotsSearchDto.getStudyId() != null) {
>  Put a comment on this code to explain the reason why we need to get all gids from studyId.

----

cc @darla-leafnode @mcrimi @clarysabel

src/main/java/org/ibp/api/rest/inventory/manager/LotResource.java
```diff
@@ -117,6 +120,9 @@
 	@Autowired
 	private GermplasmValidator germplasmValidator;
 
+	@Autowired
```
darla-leafnode, line 123: Can we remove this service from this class? Remove also the unused imports

darla-leafnode (approved):

PR #453, opened 2020-12-15
IBP-4240 import germplasm update not include codes for storage location and units

cuenyad:
Please, review


src/main/java/org/ibp/api/java/impl/middleware/germplasm/GermplasmExcelTemplateExportServiceImpl.java
```diff
@@ -189,13 +188,34 @@ private void writeObservationHeader(final HSSFSheet observationSheet, final bool
...
-			observationSheet.setColumnWidth(index, 13 * 250);
+			final String headerColumn = this.getMessageSource().getMessage(entry.getKey(), null, locale);
+			cell.setCellValue(headerColumn);
+			switch (entry.getKey()) {
+				case "export.germplasm.list.template.preferred.name.column":
+				case "export.germplasm.list.template.location.abbr.column":
+					observationSheet.setColumnWidth(index, 20 * 250);
+					break;
+				case "export.germplasm.list.template.entry.code.column":
+					observationSheet.setColumnWidth(index, 16 * 250);
```
nahuel-soldevilla, line 199: Looks ok. An improvement would be to rewrite IMPORT_HEADERS to be an enum, so that we don't repeat the string keys both here and and in the map static definition section, and we can switch over the enum

PR #471, opened 2021-02-01
IBP 4109 Update Germplasm Progenitors and Breeding Method

abatac:


src/main/java/org/ibp/api/java/impl/middleware/common/validator/GermplasmUpdateValidator.java
```diff
@@ -26,6 +30,8 @@
 @Component
 public class GermplasmUpdateValidator {
 
+	public static final String PROGENITOR_1 = "PROGENITOR 1";
```
darla-leafnode, line 33: Suggest to use the constants in Middleware so we only maintain one set

src/main/java/org/ibp/api/java/impl/middleware/germplasm/GermplasmServiceImpl.java
```diff
@@ -337,8 +337,10 @@ public long countAttributesByGid(final String gid, final List<String> attributeD
 		this.germplasmUpdateValidator.validateAttributeAndNameCodes(this.errors, germplasmUpdateDTOList);
```
darla-leafnode, line 337: I think we should abstract all this validations inside `GermplasmUpdateValidator`, which we should update name to `GermplasmUpdateDtoValidator` (similar to `germplasmImportRequestDtoValidator.validate`)

src/main/java/org/ibp/api/java/impl/middleware/common/validator/GermplasmUpdateValidator.java
```diff
@@ -135,14 +144,23 @@ public void validateLocationAbbreviation(final BindingResult errors, final Strin
...
 
-	public void validateBreedingMethod(final List<GermplasmUpdateDTO> germplasmUpdateDTOList) {
+	public void validateBreedingMethod(final BindingResult errors, final String programUUID,
+		final List<GermplasmUpdateDTO> germplasmUpdateDTOList) {
 
-		final Set<String> breedingMethodsAbbrs =
+		final List<String> breedingMethodsAbbrs =
 			germplasmUpdateDTOList.stream().filter(dto -> StringUtils.isNotEmpty(dto.getBreedingMethodAbbr()))
-				.map(dto -> dto.getBreedingMethodAbbr()).collect(Collectors.toSet());
+				.map(dto -> dto.getBreedingMethodAbbr()).collect(Collectors.toList());
```
darla-leafnode, line 152: use `GermplasmUpdateDTO::getBreedingMethodAbbr` instead of `dto -> dto.getBreedingMethodAbbr()`

src/main/java/org/ibp/api/java/impl/middleware/common/validator/GermplasmUpdateValidator.java
```diff
@@ -159,4 +177,36 @@ public void validateCreationDate(final BindingResult errors, final List<Germplas
 
 	}
 
+	public void validateProgenitorsGids(final BindingResult errors, final List<GermplasmUpdateDTO> germplasmUpdateDTOList) {
+
+		// Get the gids from progenitors.
+		final Set<Integer> progenitorGids =
+			germplasmUpdateDTOList.stream().map(dto -> dto.getProgenitors().values()).flatMap(Collection::stream)
+				.filter(value -> value != null && value != 0).collect(Collectors.toSet());
+		final List<Germplasm> germplasmByGIDs = this.germplasmMiddlewareService.getGermplasmByGIDs(new ArrayList<>(progenitorGids));
```
darla-leafnode, line 186: Can we replace usage of `this.germplasmMiddlewareService.getGermplasmByGIDs` and `this.germplasmMiddlewareService.getGermplasmByGUIDs` with `germplasmService.searchGermplasmDTO`? So that we don't need `germplasmMiddlewareService` in this class

darla-leafnode (approved):

PR #478, opened 2021-02-15
IBP-4339-ValidateFilename

gelli-leafnode:
Set directory parameter of the outpufile
Fix unit test
IBP-4339-ValidateFilename


src/main/java/org/ibp/api/rest/dataset/DatasetResource.java
```diff
@@ -348,7 +348,7 @@ private DatasetExportService getExportFileStrategy(final String fileType) {
 		final String outFileName = file.getName();
 		final HttpHeaders headers = new HttpHeaders();
 		headers
-			.add(HttpHeaders.CONTENT_DISPOSITION, String.format("attachment; filename='%s'", FileUtils.sanitizeFileName(outFileName)));
```
abatac: Why remove the quote from content disposition filename? I think we add the quote to escape the space from filename.

gelli-leafnode: Hi @abatac , 
Upon adding the single quote, the download file name includes includes single quote in file extension

darla-leafnode:
Closing this PR in favor of simpler solution

PR #480, opened 2021-02-17
IBP-4322 get locations br apiv2

mderamos:


src/main/resources/brapi/calls.json
```diff
@@ -493,6 +493,21 @@
...
+        "application/json"
+      ],
+      "datatypes": [
+        "json"
+      ],
+      "methods": [
+        "GET"
+      ],
+      "versions": [
+        "2.0"
```
darla-leafnode, line 509: Just add "2.0" as version in existing GET locations entry

darla-leafnode (approved):

PR #488, opened 2021-03-03
IBP-4392 Brapi Get Programs Improvements

abatac:
- Do not return an error if there's no program found.
- Extract crop validation to a new class.
- Return HTTP Not Impemented error if abbreviation is specified.

IBP-4392

### DIFF
--- a/src/main/java/org/ibp/api/brapi/v2/validation/CropValidator.java
+++ b/src/main/java/org/ibp/api/brapi/v2/validation/CropValidator.java
@@ -14,6 +14,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 
+// TODO: Find a centralize way to validate crop specified in URI parameter OR Request parameter.
 @Component
 public class CropValidator {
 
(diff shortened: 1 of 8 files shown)


src/main/java/org/ibp/api/brapi/v2/validation/CropValidator.java
```diff
@@ -0,0 +1,57 @@
...
+
+			if (!Util.isNullOrEmpty(userId)) {
+				final List<CropType> authorizedCrop =
+					this.workbenchDataManager.getAvailableCropsForUser(userId).stream().filter(cropType -> {
+						return cropName
+							.equalsIgnoreCase(cropType.getCropName());
+					}).collect(Collectors.toList());
+
+				if (CollectionUtils.isEmpty(authorizedCrop)) {
+					throw new ApiRequestValidationException(
```
darla-leafnode: Better to throw `AccessDeniedException`

src/main/java/org/ibp/api/brapi/v2/validation/CropValidator.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@Component
+public class CropValidator {
+
+	@Autowired
+	private WorkbenchDataManager workbenchDataManager;
```
darla-leafnode, line 22: Better to use `cropService` instead of WorkbenchDM

src/main/java/org/ibp/api/brapi/v2/validation/CropValidator.java
abatac: I used the ApiRequestValidationException to customize the message return to the client. We can't customize the message if we use the AccessDeniedException because DefaultExceptionHandler will always use the "access.denied" code message.

darla-leafnode (approved):

PR #489, opened 2021-03-05
IBP-4382 get server info implementation

mderamos:


src/main/java/org/ibp/api/java/impl/middleware/call/CallServiceImpl.java
```diff
@@ -57,10 +60,34 @@
 				brapiCalls = brapiCalls.subList(fromIndex, toIndex);
 			}
 
-			return brapiCalls;
+			return this.filterCallsByVersion(brapiCalls, version);
```
abatac, line 63: HI @mderamos , If the schema for V2 is different from v1-1.3, maybe we can just create a separate calls_v2.json file for V2? That way we don't have to filter and transform calls.json.

mderamos: Hi @abatac, @darla-leafnode and I talked about this and decided otherwise. But I think you have a point. What do you think, @darla-leafnode ?

darla-leafnode: Yes, actually it made me rethink as well. That might be better to keep the service logic simple (no filtering). Eventually, I **think** the v1 calls will be deprecated anyway.

PR #493, opened 2021-03-16
IBP-4415 Germplasm Attribute Services

mderamos:


src/main/java/org/ibp/api/java/impl/middleware/common/validator/AttributeValidator.java
```diff
@@ -27,11 +48,111 @@ public void validateAttributeIds(final BindingResult errors, final List<String>
...
+
+		if (!AttributeValidator.ALLOWED_ATTRIBUTE_TYPES.contains(attributeType.toUpperCase())) {
+			errors.reject("attribute.type.invalid", "");
+			throw new ApiRequestValidationException(errors.getAllErrors());
+		}
+	}
+
+	void validateAttributeCode(final BindingResult errors, final GermplasmAttributeRequestDto dto) {
+		final List<AttributeDTO> attributeDTOS = this.germplasmAttributeService
+			.filterGermplasmAttributes(Collections.singleton(dto.getAttributeCode()), dto.getAttributeType());
```
clarysabel, line 67: I havent tried it but does it work when passing the type in lowercase?

src/main/java/org/ibp/api/java/impl/middleware/common/validator/AttributeValidator.java
```diff
@@ -27,11 +48,111 @@ public void validateAttributeIds(final BindingResult errors, final List<String>
...
+
+		final List<GermplasmAttributeDto> existingGermplasmAttributes = germplasmAttributeDtos.stream()
+			.filter(existing -> existing.getAttributeCode().equalsIgnoreCase(dto.getAttributeCode())).collect(Collectors.toList());
+		if (!existingGermplasmAttributes.isEmpty()) {
+			errors.reject("attribute.code.invalid.existing", new String[] {dto.getAttributeCode()}, "");
+			throw new ApiRequestValidationException(errors.getAllErrors());
+		}
+	}
+
+	public void validateGermplasmAttributeExisting(final BindingResult errors, final Integer gid, final Integer attributeId) {
```
clarysabel, line 86: validateGermplasmAttributeExisting use validateGermplasmAttributeExists

src/main/java/org/ibp/api/rest/germplasm/GermplasmAttributeResource.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Set;
+
+@Api(value = "Germplasm Attribute Services")
+@Controller
+public class GermplasmAttributeResource {
+
+	@Autowired
+	private GermplasmAttributeService germplasmAttributeService;
+
+	@ApiOperation(value = "Returns germplasm attributes filtered by a list of codes and attibute type",
```
clarysabel, line 31: I think this is not the right place for this service but in the other hand, attributes will be migrated to the ontology so we will need to work on this later on

src/test/java/org/ibp/api/java/impl/middleware/common/validator/AttributeValidatorTest.java
```diff
@@ -58,11 +80,185 @@ public void testForValidProgramId() throws MiddlewareQueryException {
...
+		final List<Attribute> attributeLists = Lists.newArrayList(attribute);
 		Mockito.doReturn(attributeLists).when(this.germplasmDataManager).getAttributeByIds(Lists.newArrayList(attributeById));
 
 		this.attributeValidator.validateAttributeIds(bindingResult, Lists.newArrayList(String.valueOf(attributeById)));
 
 		Assert.assertFalse(bindingResult.hasErrors());
 	}
+
+	@Test(expected = ApiRequestValidationException.class)
+	public void testValidateAttributeType() {
```
clarysabel, line 92: In our first versions for the validator tests we were using the expected but it is much better to catch the exception so you can confirm that the error code is the right one. 
Please take a look at `GermplasmImportRequestDtoValidatorTest#testValidateBeforeSaving_ThrowsException_WhenRequestIsNull`
Also notice that your function name does not clearly express what you want to test. We not have any standard in BMS for naming our tests but see the names I have been using in GermplasmImportRequestDtoValidatorTest so you can add the condition that is tested to the function names
Please notice that each test should test only one condition. In some funtions you are testing more than one condition
This comment applies to all functions in this class so I dont repeat the same in all places

src/test/java/org/ibp/api/java/impl/middleware/germplasm/GermplasmAttributeServiceImplTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+	@Captor
+	private ArgumentCaptor<Set<String>> setArgumentCaptor;
+
+	@InjectMocks
+	private GermplasmAttributeServiceImpl germplasmAttributeServiceImpl;
+
+
+	@Test
+	public void testGetGermplasmAttributeDtos() {
```
clarysabel, line 64: same here, please give a meaningful name to the functions and verify only one condition per test

clarysabel (changes requested):

clarysabel:
HI @mderamos I have finally review both PR's. In general looks good with some exceptions in Middleware, please take care of my comments!

src/main/java/org/ibp/api/java/impl/middleware/common/validator/AttributeValidator.java
mderamos: Hi Clary, I tried and it works regardless of the case of the letters.

src/main/java/org/ibp/api/rest/germplasm/GermplasmAttributeResource.java
mderamos: Okay. I'll leave it as is for now.

PR #494, opened 2021-03-17
IBP-4428 germplasm list service

mderamos:


src/test/java/org/ibp/api/java/impl/middleware/germplasm/GermplasmListServiceImplTest.java
```diff
@@ -1356,6 +1359,20 @@ public void testAddGermplasmEntriesToList_GermplasmListIsLocked_FAIL() {
 		Mockito.verifyZeroInteractions(this.germplasmValidator);
 	}
 
+	@Test
+	public void testGetGermplasmLists() {
```
clarysabel, line 1363: Same here with the function name

clarysabel (changes requested):
HI @mderamos one comment about testing but basically the same thing about naming functions so you can apply the same to all new tests that you create in the future. In general good job with this task

PR #503, opened 2021-03-26
IBP-4358-GidWithProgenyByMID

gelli-leafnode:
Applied changes, but supplied empty list. Need only to check if germplasm has progeny


gelli-leafnode:
Closing PR. Changes only applied in Middleware

PR #506, opened 2021-04-08
IBP-4452 delete germplasm improvements

mderamos:


src/main/java/org/ibp/api/rest/germplasm/GermplasmResource.java
```diff
@@ -328,6 +328,17 @@ public long getFilteredCount() {
 		return new ResponseEntity<>(this.germplasmService.getGermplasmProgenitorDetails(gid), HttpStatus.OK);
 	}
 
+	@ApiOperation(value = "Get Germplasm used in lists other than specified.")
+	@PreAuthorize("hasAnyAuthority('ADMIN', 'CROP_MANAGEMENT', 'GERMPLASM', 'MANAGE_GERMPLASM')")
+	@RequestMapping(value = "/crops/{cropName}/germplasm/lists", method = RequestMethod.GET)
+	@ResponseBody
+	public ResponseEntity<Set<Integer>> getGermplasmPresentInOtherLists(@PathVariable final String cropName,
```
darla-leafnode, line 335: We don't a new method to check (count or get) the GIDs present in lists. Let's avoid doing API methods like this and simply include in the warning message that deletion can take a long time if they are part of any germplasm lists - from which they will also be deleted.
Please remove all code related to this, also in MW and WB.

darla-leafnode (changes requested):
Remove /crops/{cropName}/germplasm/lists endpoint and all related code. We don't want to implement API endpoints just for checking (get or count) for a warning/confirm message

src/main/java/org/ibp/api/java/impl/middleware/germplasm/GermplasmServiceImpl.java
```diff
@@ -124,6 +125,9 @@
 	@Autowired
 	private GermplasmUpdateRequestValidator germplasmUpdateRequestValidator;
 
+	@Autowired
+	private GermplasmListValidator germplasmListValidator;
```
darla-leafnode, line 129: Are we using `germplasmListValidator` in this class?

src/test/java/org/ibp/api/java/impl/middleware/germplasm/GermplasmServiceImplTest.java
```diff
@@ -72,6 +73,9 @@
 	@Mock
 	private GermplasmValidator germplasmValidator;
 
+	@Mock
+	private GermplasmListValidator germplasmListValidator;
```
darla-leafnode, line 77: Same comment, please confirm if we need `germplasmListValidator` here

darla-leafnode (approved):

PR #520, opened 2021-04-29
IBP-4474 no change password when user logged in for the first time

cuenyad:
Please, review.

### DIFF
--- a/src/main/java/org/ibp/api/domain/user/UserProfileDto.java
+++ b/src/main/java/org/ibp/api/domain/user/UserProfileDto.java
@@ -6,6 +6,9 @@ import org.pojomatic.annotations.AutoProperty;
 @AutoProperty
 public class UserProfileDto {
 
+	private String userName;
+	private String password;
+
 	private String firstName;
 	private String lastName;
 	private String email;
@@ -41,6 +44,22 @@ public class UserProfileDto {
 
 	public void setEmail(final String email) {
 		this.email = email;
+	}
+
+	public String getUserName() {
+		return this.userName;
+	}
+
+	public void setUserName(final String userName) {
+		this.userName = userName;
+	}
+
+	public String getPassword() {
+		return this.password;
+	}
+
+	public void setPassword(final String password) {
+		this.password = password;
 	}
 
 	@Override
--- a/src/main/java/org/ibp/api/domain/user/UserProfileDto.java
+++ b/src/main/java/org/ibp/api/domain/user/UserProfileDto.java
@@ -1,0 +1,61 @@
+package org.ibp.api.domain.user;
+
+import org.pojomatic.Pojomatic;
+import org.pojomatic.annotations.AutoProperty;
+
+@AutoProperty
+public class UserProfileDto {
+
+	private String firstName;
+	private String lastName;
+	private String email;
+
+	public UserProfileDto() {
+	}
+
+	public UserProfileDto(final String firstName, final String lastName, final String email) {
+		this.firstName = firstName;
+		this.lastName = lastName;
+		this.email = email;
+	}
+
+	public String getFirstName() {
+		return this.firstName;
+	}
+
+	public void setFirstName(final String firstName) {
+		this.firstName = firstName;
+	}
+
+	public String getLastName() {
+		return this.lastName;
+	}
+
+	public void setLastName(final String lastName) {
+		this.lastName = lastName;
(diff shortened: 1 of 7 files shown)


src/main/java/org/ibp/api/domain/user/UserProfileDto.java
```diff
@@ -0,0 +1,80 @@
+package org.ibp.api.domain.user;
+
+import org.pojomatic.Pojomatic;
+import org.pojomatic.annotations.AutoProperty;
+
+@AutoProperty
+public class UserProfileDto {
```
nahuel-soldevilla, line 7: Maybe `UserProfileUpdateRequestDTO` ?

src/main/java/org/ibp/api/java/impl/middleware/manager/UserValidator.java
```diff
@@ -336,15 +336,52 @@ protected void validateUserRoles(final UserDetailDto user) {
...
+	}
+
+	public void validateUserProfileLogged(final BindingResult errors, final WorkbenchUser workbenchUser){
+		final WorkbenchUser loggedInUser = this.securityService.getCurrentlyLoggedInUser();
+		if (!workbenchUser.getUserid().equals(loggedInUser.getUserid())) {
+			errors.reject("user.perfile.update.invalid.user.logged", new Object[] {}, "");
+			throw new ApiRequestValidationException(errors.getAllErrors());
+
+		}
+	}
```
nahuel-soldevilla, line 382: I know you mentioned it but I forgot. What was the reason to reeive the username  in the request and then validate it? can we omit the whole thing? just use the getCurrentlyLoggedInUser when validating the password?

src/main/java/org/ibp/api/rest/user/profile/UserProfileResource.java
```diff
@@ -0,0 +1,50 @@
...
+	@Autowired
+	private UserService userService;
+
+	@Autowired
+	private org.generationcp.middleware.service.api.user.UserService userServiceMiddleware;
+
+	@Autowired
+	private AuthenticationManager authenticationManager;
+
+	@ApiOperation(value = "Update user profile", notes = "Update First name, Laste name and Email from user profile in this deployment instance of BMSAPI. ")
```
nahuel-soldevilla, line 35: typo _Laste_

src/main/resources/messages_en.properties
```diff
@@ -423,6 +423,7 @@ users.can.not.be.edited.multiple.persons=User has duplicated person details. Con
 users.user.can.not.be.deactivated=Logged in user cannot be auto-deactivated
 users.update.superadmin.not.allowed=Updating user with superadmin role is not allowed
 
+user.perfile.update.invalid.user.logged=Update invalid, the user sent is not the same as the one logged in.
```
nahuel-soldevilla, line 426: type _perfile_, but see if we can remove the whole username validation

src/main/java/org/ibp/api/rest/user/profile/UserProfileResource.java
```diff
@@ -0,0 +1,50 @@
...
+	@RequestMapping(value = "/users/profile", method = RequestMethod.PATCH)
```
clarysabel, line 36: I will use my-profile instead

src/main/java/org/ibp/api/rest/user/profile/UserProfileResource.java
```diff
@@ -0,0 +1,50 @@
...
+	@ResponseBody
+	public ResponseEntity<Integer> updateProfile(
```
clarysabel, line 38: why do we need to return integer?

src/main/java/org/ibp/api/domain/user/UserProfileDto.java
```diff
@@ -0,0 +1,80 @@
...
+
+	private String userName;
```
clarysabel, line 9: I would not send the userName, we are changing the profile for the logged in user, so we can get the user from the session, get the username from there and validate the password

src/main/java/org/ibp/api/java/impl/middleware/manager/UserValidator.java
clarysabel: Agree, I left a similar comment in the Dto, so we dont need to send the username

PR #546, opened 2021-07-14
IBP-4724 fix samples brapi brava validator errors

mderamos:


src/main/java/org/ibp/api/domain/sample/SampleObservationDto.java
```diff
@@ -1,219 +1,159 @@
...
-
-	@Override
-	public String toString() {
-		return new ReflectionToStringBuilder(this, ToStringStyle.SHORT_PREFIX_STYLE).toString();
-	}
-}
-
+    private static final long serialVersionUID = 2340381705850740790L;
+
+    private Integer studyDbId;
```
abatac, line 15: Per brapi spec, studyDbId should be string type.

abatac (changes requested):

PR #555, opened 2021-07-21
IBP-4858 Scale type restriction for Attributes and Passport variable types.

cuenyad:
Please, review.

### DIFF
--- a/src/main/java/org/ibp/api/java/impl/middleware/ontology/validator/VariableValidator.java
+++ b/src/main/java/org/ibp/api/java/impl/middleware/ontology/validator/VariableValidator.java
@@ -399,7 +399,7 @@ public class VariableValidator extends OntologyValidator implements Validator {
 
 			final boolean isFileType = Objects.equals(scale.getDataType(), DataType.FILE_VARIABLE);
 			// Temporal restriction for Germplasm Attributes and Passport Variable Types.
-			// This restriction should be removed when implementing this ticket IBP-4798.
+			// This restriction should be removed when implementing the IBP-4798.
 			if (isFileType) {
 				final boolean isAttributeOrPassport = variable.getVariableTypes().stream().anyMatch(variableType ->
 					VARIABLE_TYPES_WITH_FILE_DATA_TYPE_RESTRICTION.contains(Integer.valueOf(variableType.getId())));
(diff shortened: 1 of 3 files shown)


src/main/java/org/ibp/api/java/impl/middleware/ontology/validator/VariableValidator.java
```diff
@@ -515,9 +535,9 @@ void variableShouldBeEditable(final VariableDetails variable, final Errors error
 			final boolean propertyEqual = Objects.equals(propertyId, oldVariable.getProperty().getId());
 			final boolean methodEqual = Objects.equals(methodId, oldVariable.getMethod().getId());
 			final boolean scaleEqual = Objects.equals(scaleId, oldVariable.getScale().getId());
-			final boolean minValuesEqual = editableVariable ? true : StringUtil.areBothEmptyOrEqual(variable.getExpectedRange().getMin(), oldVariable.getMinValue());
-			final boolean maxValuesEqual = editableVariable ? true : StringUtil.areBothEmptyOrEqual(variable.getExpectedRange().getMax(), oldVariable.getMaxValue());
-			final boolean aliasEqual = editableVariable ? true : StringUtil.areBothEmptyOrEqual(variable.getAlias(), oldVariable.getAlias());
+			final boolean minValuesEqual = editableVariable ? editableVariable : StringUtil.areBothEmptyOrEqual(variable.getExpectedRange().getMin(), oldVariable.getMinValue());
```
clarysabel, line 538: why this change is required for the file data type limitation? 
or was it a bug? if it was a bug, then should be included to be tested
@cuenyad

src/main/java/org/ibp/api/java/impl/middleware/ontology/validator/VariableValidator.java
```diff
@@ -389,6 +397,18 @@ private boolean scaleDataTypeValidationProcessor(final VariableDetails variable,
 				}
 			}
 
+			final boolean isFileType = Objects.equals(scale.getDataType(), DataType.FILE_VARIABLE);
+			// Temporally restriction for Germplasm Attributes and Passport Variable Types.
```
clarysabel, line 401: Fix typo: Temporally to temporal
Add reference to IBP-4798, this restriction should be removed when implementing this ticket.

clarysabel (changes requested):
Small changes. 
I could see some changes that are not exclusively from the restriction implementation. 
If that's the case, please edit the testing instructions and add regression testing for the scenarios that you have modified

src/main/java/org/ibp/api/java/impl/middleware/ontology/validator/VariableValidator.java
cuenyad: Yes, it's gonna be tested. 
When I added the restricction made a SonaLint cleanUp.

src/main/java/org/ibp/api/java/impl/middleware/ontology/validator/VariableValidator.java
cuenyad: done

clarysabel:
Not required because the File data type was deleted after implementing IBP-4815

PR #558, opened 2021-07-28
Ibp 3665 limit label printing results

lik-83:


src/main/java/org/ibp/api/java/impl/middleware/inventory/manager/LotServiceImpl.java
```diff
@@ -113,7 +113,12 @@
 
 	@Override
 	public List<ExtendedLotDto> searchLots(final LotsSearchDto lotsSearchDto, final Pageable pageable) {
-		return this.lotService.searchLots(lotsSearchDto, pageable);
+		return this.searchLots(lotsSearchDto, pageable, false);
+	}
+
+	@Override
+	public List<ExtendedLotDto> searchLots(final LotsSearchDto lotsSearchDto, final Pageable pageable, final boolean applyResultsLimit) {
```
clarysabel, line 120: Seems like this function is always calles with applyResultsLimit = true. Should not be better to rename the function and remove the boolean param? assuming that it will always be limited?

lik-83: Done!

PR #559, opened 2021-07-29
IBP-4791 get samples v2

mderamos:


src/main/java/org/ibp/api/brapi/v2/SampleResourceBrapi.java
```diff
@@ -0,0 +1,82 @@
+package org.ibp.api.brapi.v2;
```
abatac, line 1: Apply formatter and remove unused imports.

PR #563, opened 2021-08-03
Ibp 4839 program services

clarysabel:


src/main/java/org/ibp/api/java/impl/middleware/program/ProgramServiceImpl.java
```diff
@@ -47,7 +58,28 @@
...
+	@Autowired
+	private MethodService methodService;
+
+	@Autowired
+	private GermplasmListService germplasmListService;
+
+	private final InstallationDirectoryUtil installationDirectoryUtil = new InstallationDirectoryUtil();
+
 	public static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd");
+	public static final String UNSPECIFIED_LOCATION = "Unspecified Location";
```
nahuel-soldevilla, line 82: There's another one in org.generationcp.middleware.pojos.Location#UNSPECIFIED_LOCATION, maybe we can use that

nahuel-soldevilla (approved):

PR #569, opened 2021-08-16
IBP-4630 Added programUUID parameter to API endpoint

darla-leafnode:
Creating on @mderamos behalf

### DIFF
--- a/src/main/java/org/ibp/api/rest/keysequenceregister/KeySequenceResource.java
+++ b/src/main/java/org/ibp/api/rest/keysequenceregister/KeySequenceResource.java
@@ -26,10 +26,10 @@ public class KeySequenceResource {
 
 	@ApiOperation(value = "Delete Key Prefix Cache")
 	@PreAuthorize("hasAnyAuthority('ADMIN', 'GERMPLASM', 'MANAGE_GERMPLASM', 'DELETE_GERMPLASM')")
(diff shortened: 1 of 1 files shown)


src/main/java/org/ibp/api/rest/keysequenceregister/KeySequenceResource.java
```diff
@@ -25,10 +25,11 @@
 	private KeySequenceRegisterService keySequenceRegisterService;
 
 	@ApiOperation(value = "Delete Key Prefix Cache")
-	@PreAuthorize("hasAnyAuthority('ADMIN', 'GERMPLASM', 'MANAGE_GERMPLASM')")
-	@RequestMapping(value = "/crops/{cropName}/key-sequences", method = RequestMethod.DELETE)
+	@PreAuthorize("hasAnyAuthority('ADMIN', 'GERMPLASM', 'MANAGE_GERMPLASM', 'DELETE_GERMPLASM')")
+	@RequestMapping(value = "/crops/{cropName}/programs/{programUUID}/key-sequences", method = RequestMethod.DELETE)
```
darla-leafnode, line 29: `programUUID` should not be part of the path variable as key sequences are not program-specific resources. Better specify it as optional request parameter (same as with germplasm resources):
`@RequestParam(required = false) final String programUUID`

PR #582, opened 2021-09-02
IBP-4878 Remove AppLauncherService

cuenyad:
As a part of removing the AppLauncherService implementation.

* Removed unnecessary properties.
* Removed unnecessary tests.
* Removed methods from WorkbenchDataManager without use.

Included:
https://github.com/IntegratedBreedingPlatform/Middleware/pull/1132
https://github.com/IntegratedBreedingPlatform/Workbench/pull/545

Please, review

### DIFF
--- a/src/test/java/org/ibp/api/brapi/v2/program/ProgramResourceBrapiv2Test.java
+++ b/src/test/java/org/ibp/api/brapi/v2/program/ProgramResourceBrapiv2Test.java
@@ -205,7 +205,7 @@ public class ProgramResourceBrapiv2Test extends ApiUnitTestBase {
 
 	@Test
 	public void testListProgramFilterByAbbreviation() throws Exception {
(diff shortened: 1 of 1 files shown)


src/test/java/org/ibp/api/brapi/v2/program/ProgramResourceBrapiv2Test.java
```diff
@@ -203,20 +203,6 @@ public void testListProgramFilterByName() throws Exception {
 		;
 	}
 
-	@Test
```
clarysabel, line 206: @cuenyad 
Not clear why this test is deleted if it is about BrapI?

cuenyad: Reverted change and fixied test.

PR #595, opened 2021-09-24
IBP-4966 IBP-5048 complete breeding method services

clarysabel:


src/main/java/org/ibp/api/rest/breedingmethod/BreedingMethodResourceGroup.java
```diff
@@ -39,6 +41,37 @@
 		return new ResponseEntity<>(this.breedingMethodService.getBreedingMethod(breedingMethodDbId), HttpStatus.OK);
 	}
 
+	@ApiOperation(value = "Create breeding method")
```
clarysabel, line 44: @nahuel-soldevilla just leaving this comment because I think something is missing in the GET service. 
Say that we want to implement a view in Manage Programs where the user is able to see all breeding methods with the favorites with a star or something like that. Will the current GET implementation support it? I think,  given a programUUID and favoritesOnly=false, we dont have a way to retrieve all specifying which is favorite and which is not. 
Leaving it to check

nahuel-soldevilla: I added isFavorite field to BreedingMethodDTO, now it's available in the GET call.
I was going to add some missing filters too (by class, by name, etc) but since breeding methods is not a large dataset we can probably filter in the frontend when we implement that part.

PR #603, opened 2021-10-13
IBP-4896 IBP-5111 Add entry details to list

clarysabel:


src/main/java/org/ibp/api/java/impl/middleware/germplasm/validator/GermplasmListVariableValidator.java
```diff
@@ -0,0 +1,101 @@
...
+
+	@Autowired
+	private OntologyVariableDataManager ontologyVariableDataManager;
+
+	@Autowired
+	private GermplasmListService germplasmListService;
+
+	private BindingResult errors;
+
+	private static final List<Integer> VALID_TYPES = Lists.newArrayList(VariableType.ENTRY_DETAIL.getId());
```
nahuel-soldevilla, line 33: Will there be other allowed types for germplasm lists in the future?

PR #608, opened 2021-10-27
IBP-5094 export list data and labels

mderamos:


src/main/java/org/ibp/api/domain/common/LabelPrintingStaticField.java
```diff
@@ -62,4 +66,24 @@ public void setFieldId(final Integer fieldId) {
 		return availableStaticFieldIds;
 	}
 
+	public static LabelPrintingStaticField getByFieldId(final Integer id) {
+		if (id == null) {
+			return NONEXISTENT;
+		}
+
+		if (!LabelPrintingStaticField.STATIC_FIELD_MAP.containsKey(id)) {
```
abatac, line 74: You can simplify this by:

```
private static final Map<Integer, LabelPrintingStaticField> STATIC_FIELD_MAP_LOOKUP = new HashMap<>();
static {
		for (final LabelPrintingStaticField sf : EnumSet.allOf(LabelPrintingStaticField.class)) {
			STATIC_FIELD_MAP_LOOKUP.put(sf.getFieldId(), sf);
		}
	}
```
```
public static Optional<LabelPrintingStaticField> getByFieldId(final Integer id) {
		if (id == null) {
			return Optional.empty();
		}
		if (LabelPrintingStaticField.STATIC_FIELD_MAP_LOOKUP.containsKey(id)) {
			return Optional.of(STATIC_FIELD_MAP_LOOKUP.get(id));
		}
		return Optional.empty();

	}
```
Use optional instead of creating enum for non existent field.

src/main/java/org/ibp/api/rest/labelprinting/GermplasmLabelPrinting.java
```diff
@@ -158,43 +159,48 @@ public LabelsNeededSummaryResponse transformLabelsNeededSummary(
 
 	@Override
 	public OriginResourceMetadata getOriginResourceMetadata(
-		final LabelsInfoInput labelsInfoInput) {
+		final LabelsInfoInput labelsInfoInput, final String programUUID) {
```
abatac, line 162: Remove this unused parameter

PR #611, opened 2021-11-01
IBP-5093 IBP-5088 import list with entry details and entry details updates

nahuel-soldevilla:


src/main/java/org/ibp/api/java/impl/middleware/germplasm/GermplasmListServiceImpl.java
```diff
@@ -467,6 +487,61 @@ private void processEntries(final GermplasmListGeneratorDTO request, final Strin
...
+
+			entryDetail.setVariableId(variable.getId());
+			entryDetail.setcValueId(VariableValueUtil.resolveCategoricalValueId(variable, value));
+		}
+	}
+
+	private void validateVariableDataTypeValue(final Variable variable, final String value) {
+		if (!VariableValueUtil.isValidAttributeValue(variable, value)) {
+			throw new ApiRequestValidationException("invalid.variable.value.with.param", new String[] {
+				"variable: " + variable.getName() + ", value: " + value
```
clarysabel, line 539: to be improved, I suggest we have a different property for this scenario, basically variable and value are hardcoded in english

src/main/java/org/ibp/api/java/impl/middleware/germplasm/GermplasmListServiceImpl.java
```diff
@@ -467,6 +487,61 @@ private void processEntries(final GermplasmListGeneratorDTO request, final Strin
...
+			});
+		}
+		checkArgument(value.length() <= 255, "text.field.max.length", new String[] {"entry detail value", "255"});
```
clarysabel, line 542: same here, in case we want to reuse the property I think we can use the attribute name instead of a custom text so at least it is associated the the body we are expecting

src/main/resources/messages_en.properties
```diff
@@ -459,6 +461,9 @@ error.germplasmlist.save.gid=gid is mandatory
 error.germplasmlist.save.gid.not.exists=gid {0} does not exists
 error.germplasmlist.save.type.not.exists=list type {0} does not exist
 
+error.germplasmlist.importupdates.entryno.mandatory=entryNo is mandatory
+error.germplasmlist.importupdates.listid.mandatory=list id is mandatory
```
clarysabel, line 465: should this be listId instead of list id?

PR #612, opened 2021-11-04
Ibp 5172 deprecate location.program UUID

lik-83:


src/main/java/org/ibp/api/java/impl/middleware/program/validator/ProgramFavoriteValidator.java
```diff
@@ -67,7 +67,7 @@ public void validateAddFavorites(final String programUUID, final ProgramFavorite
 
 		switch (favoriteType) {
 			case LOCATION:
-				this.validateLocationId(errors, programUUID, entityIds);
+				this.validateLocationId(errors, entityIds);
 				break;
 			case METHOD:
 				this.validateMethodId(errors, programUUID, entityIds);
```
clarysabel, line 73: in validateMethodId, programUUID is not used @lik-83 could you please remove the parameter?

clarysabel (approved):

src/main/java/org/ibp/api/java/impl/middleware/program/validator/ProgramFavoriteValidator.java
lik-83: Done!

PR #630, opened 2021-12-06
IBP-5273 Enable Lock/Unlock List If User Has Full Access Or User is the Owner

abatac:


src/test/java/org/ibp/api/java/impl/middleware/germplasm/GermplasmListServiceImplTest.java
```diff
@@ -1650,6 +1658,28 @@ public void toggleGermplasmListStatus_notOwner() {
 		Mockito.verify(this.securityService).getCurrentlyLoggedInUser();
 	}
 
+	@Test
+	public void toggleGermplasmListStatus_notOwner_userIsAdmin() {
```
darla-leafnode, line 1662: rename "userIsAdmin" to "userhasAdminPermission"

darla-leafnode (approved):

PR #690, opened 2022-04-22
IBP-5612 - add validation on germplasm import update for when gid equa…

cheryleigh:
…ls a progenitor

### DIFF
--- a/src/main/java/org/ibp/api/java/impl/middleware/common/validator/GermplasmUpdateDtoValidator.java
+++ b/src/main/java/org/ibp/api/java/impl/middleware/common/validator/GermplasmUpdateDtoValidator.java
@@ -58,16 +58,21 @@ public class GermplasmUpdateDtoValidator {
 	public void validate(final String programUUID, final List<GermplasmUpdateDTO> germplasmUpdateDTOList) {
 
 		final BindingResult errors = new MapBindingResult(new HashMap<>(), GermplasmUpdateDTO.class.getName());
+		final Set<String> germplasmUUIDs =
+			germplasmUpdateDTOList.stream().map(dto -> StringUtils.isNotEmpty(dto.getGermplasmUUID()) ? dto.getGermplasmUUID() : null)
+				.filter(
+					Objects::nonNull).collect(Collectors.toSet());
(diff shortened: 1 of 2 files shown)


src/main/java/org/ibp/api/java/impl/middleware/common/validator/GermplasmUpdateDtoValidator.java
```diff
@@ -194,6 +195,15 @@ protected void validateAttributeAndNameCodes(final BindingResult errors, final S
 		}
 	}
 
+	protected void validateProgenitorMustNotBeGid(final BindingResult errors, final List<GermplasmUpdateDTO> germplasmUpdateDTOList) {
```
clarysabel, line 198: @cheryleigh Please use a more self explanatory name for the function
Does this work when excel file contains GUID insted of GID ?

src/main/java/org/ibp/api/java/impl/middleware/common/validator/GermplasmUpdateDtoValidator.java
```diff
@@ -194,6 +195,25 @@ protected void validateAttributeAndNameCodes(final BindingResult errors, final S
 		}
 	}
 
+	protected void validateCannotAssignSelfAsProgenitor(final BindingResult errors, final List<GermplasmUpdateDTO> germplasmUpdateDTOList) {
```
clarysabel, line 198: Hi @cheryleigh 
My concern is we are going 2 times to query by GUID (see this.validateGermplasmIdAndGermplasmUUID) 
I think we can split `validateGermplasmIdAndGermplasmUUID` and return the list og germplasm as optional and use that list to run a 2nd validation 
I dont see any risk on doing this improvement now. If you find it complicated, let me know and we can create a follow up task

cheryleigh: Hi @clarysabel ,

I updated the implementation to retrieve the list first and pass it as parameter to the 2 methods. Kindly check if this is alright. Thank you!

PR #691, opened 2022-04-22
IBP-5590-Limit-add-column-in-germplasm-list

cuenyad:
Please, review

Included:
https://github.com/IntegratedBreedingPlatform/Workbench/pull/692
https://github.com/IntegratedBreedingPlatform/Middleware/pull/1258


src/main/java/org/ibp/api/java/impl/middleware/common/validator/GermplasmListValidator.java
```diff
@@ -150,6 +154,17 @@ public void validateListIsUnlocked(final GermplasmList germplasmList) {
 		}
 	}
 
+	public void validateExceededMaximunColumnsAllowed(final GermplasmList germplasmList ,final List<GermplasmListDataUpdateViewDTO> view) {
```
clarysabel, line 157: rename to validateMaxColumnsAllowed

src/main/java/org/ibp/api/java/impl/middleware/common/validator/GermplasmListValidator.java
```diff
@@ -150,6 +154,17 @@ public void validateListIsUnlocked(final GermplasmList germplasmList) {
...
+		final int entryDetailsLength = germplasmList.getView().stream().filter(germplasmListDataView -> germplasmListDataView.isEntryDetailColumn()).collect(
```
clarysabel, line 158: @cuenyad I think we can do the same with variables to prevent this problem to be caused by entry variables. You can move the count search to middleware and sum both props and columns and make the proper validation in both services in BMSAPI

clarysabel (changes requested):

src/main/java/org/ibp/api/java/impl/middleware/common/validator/GermplasmListValidator.java
```diff
@@ -150,6 +152,21 @@ public void validateListIsUnlocked(final GermplasmList germplasmList) {
 		}
 	}
 
+	public void validateMaxColumnsAllowed(final Integer listId) {
```
clarysabel, line 155: unify both functions sending the property as a parameter @cuenyad

src/main/resources/messages_en.properties
```diff
@@ -658,6 +658,8 @@ list.id.required=ListId is required
 list.id.invalid=We have not found the list id {0} in the system
 list.invalid=ListId is not a list
 list.locked=List is locked
+list.add.columns.exceeded.maximun.allowed=The number of columns add exceeded the maximum allowed.
```
clarysabel, line 661: The number of added columns exceeds the allowed max of 50. Either remove entry variables or columns to proceed

src/main/resources/messages_en.properties
```diff
@@ -658,6 +658,8 @@ list.id.required=ListId is required
...
+list.add.variables.exceeded.maximun.allowed=The number of variables add exceeded the maximum allowed. For add more you should remove some columns.
```
clarysabel, line 662: The number of added variables exceeds the allowed max of 50. Either remove entry variables or columns to proceed

PR #698, opened 2022-05-05
IBP-5629- Variable Resource Changes to support Means Data Upload

abatac:


src/main/java/org/ibp/api/java/impl/middleware/ontology/brapi/VariableServiceBrapiImpl.java
```diff
@@ -42,12 +47,37 @@ public long countObservationVariables(final VariableSearchRequestDTO requestDTO)
...
+			response.setErrors(bindingResult.getAllErrors());
+		}
+
+		final List<VariableDTO> variables = this.middlewareVariableServiceBrapi.createObservationVariables(variableDTOList);
+		if (CollectionUtils.isNotEmpty(variables)) {
+			final VariableSearchRequestDTO variableSearchRequestDTO = new VariableSearchRequestDTO();
+			// Fetch the saved observation variables from the database to populate the fields.
+			variableSearchRequestDTO.setObservationVariableDbIds(variables.stream().map(VariableDTO::getObservationVariableDbId).collect(
+				Collectors.toList()));
+			final List<VariableDTO> createdVariables = this.getObservationVariables(crop, variableSearchRequestDTO, null);
```
clarysabel, line 77: @abatac is there any way to avoid querying variables if the createObservationVariables is already returning the list of VariableDTO? Maybe completing attributes that are missing in middleware will help us to avoid this extra call to get

abatac: Hi Clary, I've fixed this. I moved the logic to retrieve the variableDTOs to make sure the fields are populated directly from middleware VariableServiceBrapiImpl#createObservationVariables.

PR #702, opened 2022-05-09
IBP-4985_remove_single_entity_response

cuenyad:
Please, review

Included:
https://github.com/IntegratedBreedingPlatform/Fieldbook/pull/1081
https://github.com/IntegratedBreedingPlatform/InventoryManager/pull/102
https://github.com/IntegratedBreedingPlatform/Workbench/pull/705


clarysabel (commented):
Hi @cuenyad just trying to remember/understand why we are returning SearchDto instead of String. Is there any reason for that? If not, we can just respond with String. 
Let's decide it before merging because a new change will also affect inventory 
To discuss later on monday

nahuel-soldevilla:
> Hi @cuenyad just trying to remember/understand why we are returning SearchDto instead of String. Is there any reason for that? If not, we can just respond with String. Let's decide it before merging because a new change will also affect inventory To discuss later on monday

This is probably because of angularjs http not being able to deal correctly with primitives in response, I remember having to wrap some strings in a map just for this reason: https://github.com/IntegratedBreedingPlatform/BMSAPI/blob/5a67ac5b54514079a2485c1188a4123b2ea35884/src/main/java/org/ibp/api/rest/file/FileResource.java#L118

cc @cuenyad

PR #707, opened 2022-05-24
IBP-3796: handle regeneration of fieldmap

clarysabel:
- transfer delete fieldmap API from fieldbook to bmsapi


src/main/java/org/ibp/api/rest/study/StudyResource.java
```diff
@@ -160,4 +169,28 @@ public ResponseEntity<List<StudyDTO>> getStudies(
...
+	/**
+	 * Delete field map.
+	 *
+	 * @param form the form
+	 * @param model the model
+	 * @return the string
+	 */
+	@ApiOperation(value = "Delete specified fieldmaps of study", notes = "Delete specified fieldmaps of study")
+	@ResponseBody
+	@RequestMapping(value = "/{cropName}/programs/{programUUID}/studies/{studyId}/deleteFieldmap/{ids}", method = RequestMethod.DELETE)
```
clarysabel, line 181: @cheryleigh deleting multiple resources with a single REST API is not clearly solved by the specification. 
I think sending a list of items separated by `,` in a DELETE request is confusing, it can be interpreted as it is a single resource with a string id that contains `,` 
I suggest to convert it to POST /{cropName}/programs/{programUUID}/studies/{studyId}/fieldmaps/deletion and send the list of ids in the request body

PR #710, opened 2022-06-03
IBP-5714 storage location default

mderamos:


src/main/java/org/ibp/api/java/impl/middleware/common/validator/LocationValidator.java
```diff
@@ -229,16 +229,21 @@ public void validateCanBeDeleted(final Integer locationId) {
 	}
 
 	private void validateLocationNotDeletable(final LocationDTO locationDTO) {
-		if (LOCATIONS_NOT_DELETABLES.contains(locationDTO.getName()) || locationDTO.isDefaultLocation()) {
+		if (LOCATIONS_NOT_DELETABLES.contains(locationDTO.getName())) {
```
clarysabel, line 232: @mderamos @mcrimi 
Does validating that the following locations can not be deleted still make sense : "Unspecified Location", "Default Seed Store", "Default Breeding Location"? 
With the new implementation seems that they can be removed if in any instance they are not used as default anymore

src/main/java/org/ibp/api/java/location/LocationService.java
```diff
@@ -12,7 +12,9 @@ public interface LocationService {
 
 	LocationDTO getLocation(Integer locationId);
 
-	LocationDTO getDefaultLocation(String programUUID);
+	LocationDTO getBreedingLocationDefault(String programUUID);
```
clarysabel, line 15: rename to getDefaultBreedingLocation

src/main/java/org/ibp/api/java/location/LocationService.java
```diff
@@ -12,7 +12,9 @@ public interface LocationService {
...
+
+	LocationDTO getStorageLocationDefault(String programUUID);
```
clarysabel, line 17: rename to getDefaultStorageLocation

src/main/java/org/ibp/api/rest/location/LocationResource.java
```diff
@@ -51,13 +51,22 @@ public ResponseEntity<LocationDTO> getLocation(
 		return new ResponseEntity<>(this.locationService.getLocation(locationId), HttpStatus.OK);
 	}
 
-	@ApiOperation(value = "Get default location")
-	@RequestMapping(value = "/crops/{cropName}/programs/{programUUID}/locations/default", method = RequestMethod.GET)
+	@ApiOperation(value = "Get breeding location default")
+	@RequestMapping(value = "/crops/{cropName}/programs/{programUUID}/locations/breeding-location-default", method = RequestMethod.GET)
```
clarysabel, line 55: If I have to implement it I will leave as it was (/default) and send a type by param to the GET function so we have only one default service per program 
@mderamos

src/main/java/org/ibp/api/java/impl/middleware/common/validator/LocationValidator.java
mderamos: Hi @clarysabel, I agree. Also setting the Unspecified Location as program favorite upon program creation seems unnecessary since the users are selecting default locations and setting the selected locations as favorites.

mcrimi: Hi @mderamos, yes, that sounds reasonable to me. cc: @clarysabel

PR #725, opened 2022-07-12
IBP-5767: implement filtering lot by attributes

cheryleigh:


src/main/java/org/ibp/api/rest/germplasm/GermplasmResource.java
```diff
@@ -185,7 +187,8 @@ public ResponseEntity<List<Variable>> searchAttributes(@PathVariable final Strin
 		@RequestParam(required = false) final String programUUID,
 		@RequestParam(required = true) final String query) {
 
-		return new ResponseEntity<>(this.variableService.searchAttributeVariables(query, programUUID), HttpStatus.OK);
+		return new ResponseEntity<>(this.variableService.searchAttributeVariables(query,
```
clarysabel, line 190: @cheryleigh seems we are duplicating this service, adding a new one in VariableResource, that internally are doing the same 
I suggest we remove this one and just leave the one in VariableResource, with the endpoint in VariableResource. 
It does not make much sense to have /germplasm in the endpoint 
You will have to change the frontend to call the service in VariableService.

PR #730, opened 2022-07-29
IBP-5770: implement import entry details in germplasm and checks

cheryleigh:


src/main/java/org/ibp/api/domain/study/StudyEntryDetailsImportRequest.java
```diff
@@ -0,0 +1,40 @@
+package org.ibp.api.domain.study;
+
+import org.ibp.api.domain.dataset.DatasetVariable;
+
+import java.util.List;
+
+public class StudyEntryDetailsImportRequest {
+
+	private List<StudyEntryDetailsValueMap> data;
```
abatac, line 9: Implement hashCode, equals and toString

src/main/java/org/ibp/api/domain/study/StudyEntryDetailsValueMap.java
```diff
@@ -0,0 +1,38 @@
+package org.ibp.api.domain.study;
+
+import org.generationcp.middleware.service.api.dataset.StockPropertyData;
+
+import java.util.List;
+
+public class StudyEntryDetailsValueMap {
+
+	private String entryNumber;
```
abatac, line 9: Implement hashCode, equals and toString

src/main/java/org/ibp/api/java/impl/middleware/dataset/DatasetServiceImpl.java
```diff
@@ -181,6 +182,25 @@ public long countObservationsByInstance(final Integer studyId, final Integer dat
...
+
+	@Override
+	public void addDatasetVariables(final Integer studyId, final List<DatasetVariable> datasetVariables) {
+		this.studyValidator.validate(studyId, true);
+
+		final Integer datasetId = this.getDatasets(
+			studyId, Collections.singleton(DatasetTypeEnum.PLOT_DATA.getId())).get(0).getDatasetId();
+
+		datasetVariables.forEach(datasetVariable -> {
+			this.getMeasurementVariable(studyId, datasetId, datasetVariable);
```
abatac, line 198: It seems that MeasurementVariable is not necessary for this new method. Refactor this to avoid it.

src/main/java/org/ibp/api/java/impl/middleware/dataset/DatasetServiceImpl.java
```diff
@@ -181,6 +182,25 @@ public long countObservationsByInstance(final Integer studyId, final Integer dat
...
+
+		return this.getMeasurementVariable(studyId, datasetId, datasetVariable);
+
+	}
+
+	@Override
+	public void addDatasetVariables(final Integer studyId, final List<DatasetVariable> datasetVariables) {
+		this.studyValidator.validate(studyId, true);
+
+		final Integer datasetId = this.getDatasets(
```
abatac, line 194: Make the datasetId as a method parameter so that this can be reused with other dataset.

abatac (changes requested):

src/main/java/org/ibp/api/java/impl/middleware/study/StudyEntryServiceImpl.java
```diff
@@ -302,6 +328,70 @@ public List<StudyEntryColumnDTO> getStudyEntryColumns(final Integer studyId, fin
...
+
+	private void processStudyEntry(final Integer studyId, final List<StockPropertyData> entriesDataList,
+		final StudyEntryDto entry) {
+		if (!org.fest.util.Collections.isEmpty(entriesDataList)) {
+			entriesDataList.forEach(variable -> {
+				variable.setStockId(entry.getEntryId());
+				if (variable.hasValue() && !variable.getValue().isEmpty()) {
+					this.studyEntryObservationService.createOrUpdateObservation(studyId, variable);
+				} else {
+					final Optional<StockProperty> stockPropertyOptional = this.middlewareStudyEntryService.getByStockIdAndTypeId(
```
abatac, line 384: To improve performance, I think it's better to retrieve the stock properties in one go rather than querying them one by one. I would convert the stock properties to MultiKeyMap (stockid and typeid as keys) and use that to check if a record with stockid+typeid combination exists.

abatac (approved):

PR #737, opened 2022-08-03
IBP-5773 Delete Georeference

abatac:


src/main/java/org/ibp/api/rest/study/StudyInstanceResource.java
```diff
@@ -137,4 +139,15 @@ public ResponseEntity<InstanceDescriptorData> updateInstanceDescriptor(final @Pa
 			HttpStatus.OK);
 	}
 
+	@ApiOperation(value = "Delete georeference associated to an instance",
+		notes = "Delete georeference of associated to an instance")
+	@PreAuthorize("hasAnyAuthority('ADMIN','STUDIES','MANAGE_STUDIES', 'BROWSE_STUDIES')")
```
clarysabel, line 144: @abatac is BROWSE_STUDIES needed? or just because it was copied and pasted 
Seems that it is not

abatac: @clarysabel, yup BROWSE_STUDIES is not needed, I'll remove it.

abatac: Done.

PR #743, opened 2022-08-12
IBP-5835-ontology_method_for_sta_brapp

cuenyad:
Please, review

Included:
https://github.com/IntegratedBreedingPlatform/Workbench/pull/744
https://github.com/IntegratedBreedingPlatform/Middleware/pull/1328

### DIFF
--- a/src/main/java/org/ibp/api/domain/ontology/Usage.java
+++ b/src/main/java/org/ibp/api/domain/ontology/Usage.java
@@ -33,7 +33,7 @@ public class Usage {
 	// lists of variable
 	private Integer lists;
 
-	private boolean system;
+	private boolean systemTerm;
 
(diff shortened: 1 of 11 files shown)


src/main/java/org/ibp/api/domain/ontology/Usage.java
```diff
@@ -33,7 +33,7 @@ public class Usage {
 	// lists of variable
 	private Integer lists;
 
-	private boolean isSystemVariable;
+	private boolean system;
```
clarysabel, line 36: rename to isSystemTerm @cuenyad

cuenyad: done.

PR #756, opened 2022-09-15
IBP-5870-edit-entry-details-values-when-the-observation-units-are-generated

lik-83:


src/main/java/org/ibp/api/java/impl/middleware/dataset/validator/DatasetValidator.java
```diff
@@ -232,4 +235,33 @@ public void validateVariableBelongsToVariableType(final int datasetId, final int
 			throw new ApiRequestValidationException(this.errors.getAllErrors());
 		}
 	}
+
+	public void validateNotDeletingSystemEntryDetailToAlreadyGeneratedExperiment(final Integer datasetId,
```
clarysabel, line 239: @lik-83 just a suggestion, I would use Removing instead of Deleting, same in messages

lik-83: Done!

PR #757, opened 2022-09-16
IBP-5907 implement post v2 images

mderamos:


src/main/java/org/ibp/api/java/impl/middleware/file/validator/FileValidator.java
```diff
@@ -2,36 +2,84 @@
...
+			this.errors.reject("file.upload.not-supported",  new String[] {StringUtils.join(this.supportedFileTypes, ", ")}, "");
+			throw new ApiRequestValidationException(this.errors.getAllErrors());
 		}
 	}
 
+	public void validateFileStorage() {
+		BaseValidator.checkArgument(this.fileStorageService.isConfigured(), "file.storage.not.configured");
+	}
+
+	public void validateImage(final ImageNewRequest body) {
```
clarysabel, line 58: @mderamos this function is very specific for BRAPI. It will me much clear to keep FileValidator BrAPI agnostic and move this validateImage to a new validator class or to the resource

src/main/resources/messages_en.properties
```diff
@@ -1041,6 +1041,9 @@ file.upload.brapi.images.filename.required=imageFileName is required
 file.upload.brapi.images.observationunitdbid.required=ObservationUnitDbId is required
 file.upload.entity.invalid=Only one of the following parameters should be provided: observationUnitUUID, germplasmUUID, instanceId, or lotId should be provided.
 file.upload.detach.parameters.invalid=Only one of the following parameters should be provided: observationUnitUUID, germplasmUUID, instanceId, or lotId should be provided.
+file.upload.reference.null=The request is invalid because some external reference information is either null or empty.
+file.upload.reference.id.exceeded.length=The request is invalid because referenceID must not exceed length of 2000.
```
clarysabel, line 1045: @mderamos I think we are not using "The request is invalid because" in our messages in general. 
the response code is already informing that is invalid, I think we can remove that from both messages

PR #762, opened 2022-10-05
IBP-2676: implement handling of obsolete variables

cheryleigh:


src/main/java/org/ibp/api/rest/ontology/VariableResource.java
```diff
@@ -138,13 +138,15 @@ public ResponseEntity<List<VariableDetails>> listAllVariablesUsingFilter(
 			+ " to retrieve possible property class values that can be supplied here as a comma separated list.")
 		@RequestParam(value = "propertyClasses", required = false) final Set<String> propertyClasses,
 
+		@RequestParam(value = "showObsoletes", defaultValue = "true") final boolean showObsoletes,
```
abatac, line 141: I think it's better to set the default value for showObsoletes to false.

abatac (changes requested):

PR #768, opened 2022-10-25
IBP-5978-RestrictNumberOfActiveUsers

mderamos:
IBP-5978


src/main/java/org/ibp/api/java/impl/middleware/manager/UserValidator.java
```diff
@@ -379,6 +388,12 @@ public void validateEmail(final BindingResult errors, final String email) {
 		this.validatePersonEmailIfExists(errors, email);
 	}
 
+	public void validateNumberOfActiveUsers(final BindingResult errors) {
+		if ((int)this.userService.countAllActiveUsers() >= this.maximumActiveUsers) {
```
clarysabel, line 392: @mderamos is this (int) cast really needed? I think we dont

PR #771, opened 2022-11-04
IBP-5874 Implement Two Factor Authentication

abatac:


src/main/java/org/ibp/api/domain/user/UserDetailDto.java
```diff
@@ -0,0 +1,160 @@
...
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.generationcp.middleware.domain.workbench.CropDto;
+import org.generationcp.middleware.service.api.user.UserRoleDto;
+
+import java.io.Serializable;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class UserDetailDto implements Serializable, Comparable<UserDetailDto> {
```
clarysabel, line 12: @abatac UserDetailDto should not exist anymore, use UserDto in Middleware instead.

PR #772, opened 2022-11-07
IBP-5977: implement design license validation on login 

cheryleigh:


src/main/java/org/ibp/api/ibpworkbench/rest/BreedingView.java
```diff
@@ -117,5 +140,44 @@ public void setBreedingViewService(final BreedingViewService breedingViewService
...
 
+	@ApiOperation(value = "Gets list of Breeding View Licenses")
+	@RequestMapping(value = "/validate-license", method = RequestMethod.POST)
+	@ResponseBody
+	public ResponseEntity<Map<String, Object>> validateBVLicense() {
+		final Map<String, Object> out = new LinkedHashMap<>();
+		HttpStatus isSuccess = HttpStatus.BAD_REQUEST;
+
+		try {
+			final License designLicenseInfo = this.designLicenseService.getLicenseInfo();
```
abatac, line 151: There's already an existing API to get the BV License. Please see BVLicenseResource (GET /breeding-view-licenses). You can use that api to get the license data directly from front-end, and just implement this validation also in the front-end.

src/main/resources/application.properties
```diff
@@ -42,6 +42,7 @@ upload.directory=${upload.directory}
 bv.design.path=${bv.design.path}
 # Expected time unit is in minutes
 bv.design.runner.timeout=${bv.design.runner.timeout}
+bv.design.validation.on.login.enabled=${bv.design.validation.on.login.enabled}
```
abatac, line 45: I think you can remove this change.

PR #794, opened 2023-01-09
IBP-5984-implement-post-advance-samples

lik-83:


src/main/java/org/ibp/api/rest/dataset/DatasetResource.java
```diff
@@ -545,6 +544,19 @@ public ResponseEntity<Void> updatePlotDatasetProperties(@PathVariable final Stri
 		return new ResponseEntity<>(HttpStatus.OK);
 	}
 
+	@ApiOperation( value = "Returns the variables associated to the given study/dataset filtered by a given variable types")
+	@PreAuthorize("hasAnyAuthority('ADMIN','STUDIES', 'MANAGE_STUDIES')")
+	@RequestMapping(value = "/{crop}/programs/{programUUID}/studies/{studyId}/variables/types/{variableTypeIds}", method = RequestMethod.GET)
```
clarysabel, line 549: sounds like variableTypeIds  should be a RequestParam instead of PathVariable, removing types from the URL and specially because we have an unrequired parameter datasetId that represents a higher level than the variable type

clarysabel (approved):

PR #796, opened 2023-01-13
IBP-6160-error_exporting_germplasm_list

cuenyad:
Please, review


nahuel-soldevilla (commented):
As we discuss, this looks fine to me, although the perf issue in germplasmService.searchGermplasm (with the cross expansion service) still exists, we are just bypassing it here. We'll need to address it the future.

cc @cuenyad @clarysabel

PR #809, opened 2023-02-28
IBP-5924 granular permissions to MG Studies

cuenyad:
Please, review

Included:
https://github.com/IntegratedBreedingPlatform/Middleware/pull/1430
https://github.com/IntegratedBreedingPlatform/Workbench/pull/829
https://github.com/IntegratedBreedingPlatform/Fieldbook/pull/1164


src/main/java/org/ibp/api/rest/dataset/DatasetResource.java
```diff
@@ -120,7 +120,9 @@ public ResponseEntity<String> countPhenotypes(
 	}
 
 	@ApiOperation(value = "Add Dataset Variable", notes = "Add Dataset Variable")
-	@PreAuthorize("hasAnyAuthority('ADMIN','STUDIES', 'MANAGE_STUDIES')")
+	@PreAuthorize("hasAnyAuthority('ADMIN','STUDIES', 'MANAGE_STUDIES','CREATE_STUDIES', 'DELETE_STUDY', 'CLOSE_STUDY', 'LOCK_STUDY','MS_MANAGE_OBSERVATION_UNITS','MS_WITHDRAW_INVENTORY','MS_CREATE_PENDING_WITHDRAWALS', "
```
clarysabel, line 123: @cuenyad not following why in many cases we have added a lot of permissions that seems to be not related with the service, for example why someone that has permissions to close a study, has permissions to add a variable to the study?

cuenyad: This comment does not apply. The service was removed.

